PR #106, opened 2015-02-20
Support 'ms' in _datetime_to_epoch

### LINKED ISSUE #105: DataFrameClient time_precision='ms' does not work
Specifying time_precision='ms' to e.g. write_points method results in an error
`400: time field must be float but is <nil> (<nil>)`
Using time_precision='m' works.

ReAzem:
Closes #105 

//cc @timtroendle @malaakso


timtroendle:
Looks good to me!

ReAzem:
Thanks!

PR #113, opened 2015-03-04
added default database to InfluxDBClient query method

nemesisdesign:
Avoid having to specify the database at each query

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -211,6 +211,8 @@ class InfluxDBClient(object):
         params['q'] = query
         if database:
             params['db'] = database
+        else:
+            params['db'] = self._database
 
         response = self.request(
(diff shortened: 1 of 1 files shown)


ReAzem:
@nemesisdesign I think we should do this in the functions calling ```query``` instead, because there are queries that dont require a database argument (like get_list_series).

For example, I do it [here](https://github.com/nemesisdesign/influxdb-python/blob/patch-1/influxdb/client.py#L381)

nemesisdesign:
ok, 1 OT question, is _query dead code from the old version, or is it planned to be implemented or what is it for?

ReAzem:
@nemesisdesign 
Its dead code. We should remove it. Do you want to open an PR or should I do it?

PR #115, opened 2015-03-07
Fix example server_data to support v0.9

areski:
This includes fixes for the private method '_query' from client.py to work
properly with influxdb v0.9 API

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -297,7 +297,6 @@ class InfluxDBClient(object):
         if self.use_udp:
             self.send_packet(data)
         else:
-            import ipdb; ipdb.set_trace()
             self.write(
                 data=data,
                 expected_response_code=200
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -195,50 +195,39 @@ class InfluxDBClient(object):
               query,
               params={},
               expected_response_code=200,
+              time_precision='s',
               database=None,
-              raw=False):
+              raw=False,
+              chunked=False):
         """
         Query data
 
         :param params: Additional parameters to be passed to requests.
+        :param time_precision: [Optional, default 's'] Either 's', 'm', 'ms'
+            or 'u'.
         :param database: Database to query, default to None.
         :param expected_response_code: Expected response code. Defaults to 200.
         :param raw: Wether or not to return the raw influxdb response.
+        :param chunked: [Optional, default=False] True if the data shall be
+            retrieved in chunks, False otherwise.
+
         """
+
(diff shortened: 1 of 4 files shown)


ReAzem:
_query is no longer used. Changed should be made in the ```query``` method instead.

ReAzem:
Keep in mind that you can no longer query with a specific precision:

> The format of the returned timestamps complies with RFC3339, and has nanosecond precision.

areski:
I like the idea of not having to set database all the time, maybe we should open a new ticket to follow up on this.

areski:
See last commit, it removes the query method and make query method compliant with dataframe_client

influxdb/client.py
```diff
@@ -276,6 +297,7 @@ def _write_points(self,
         if self.use_udp:
             self.send_packet(data)
         else:
+            import ipdb; ipdb.set_trace()
```
ReAzem: @areski  oops!

areski:
I often have the bad habbits to count on travis-ci to run the tests :)
debugging removed! thanks

influxdb/client.py
```diff
@@ -195,21 +195,40 @@ def query(self,
...
 
+        if time_precision not in ['s', 'm', 'ms', 'u']:
+            raise Exception(
+                "Invalid time precision is given. (use 's', 'm', 'ms' or 'u')")
+
         params['q'] = query
         if database:
             params['db'] = database
 
+        if time_precision:
```
ReAzem: This condition is useless if we default the time precision to 's'. I would prefer to default the precision to None so that we use the InfluxDB default. I'll fix it myself when I merge.

influxdb/client.py
```diff
@@ -195,21 +195,40 @@ def query(self,
               query,
               params={},
               expected_response_code=200,
+              time_precision='s',
```
ReAzem: There is no time precision in queries.

see: http://influxdb.com/docs/v0.9/concepts/reading_and_writing_data.html

> The format of the returned timestamps complies with RFC3339, and has nanosecond precision

influxdb/client.py
```diff
@@ -195,21 +195,40 @@ def query(self,
...
               database=None,
-              raw=False):
+              raw=False,
+              chunked=False):
```
ReAzem: Chunked HTTP responses were dropped. They are no longer in the 0.9 docs.

ReAzem:
I'll take care of it.

areski:
Ok I'm leaving this to you

ReAzem:
@areski Thanks =)

PR #118, opened 2015-03-09
handle error nicely when trying to import pandas

areski:


otoolep:
I don't see the difference, either way an exception is raised.

areski:
There is no difference, just keeping the import in the header and make sure that pd is defined.  
This is an estetic fix

ReAzem:
Thanks!

PR #124, opened 2015-03-14
close response objects

jove1:
Hi,

requests normally reuses the tcp connection for multiple requests, this is handy when feeding data to influx continuously. However to reuse the connection, previous response object has to be closed, this happens implicitly when its content is accessed (.text, .content, .json()). However, it has to be done manually when the data is ignored.

Don't know if this is the best way to patch it, though.

http://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow
"Requests cannot release the connection back to the pool unless you consume all the data or call Response.close."

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -181,7 +181,7 @@ class InfluxDBClient(object):
             params=None,
             data=data,
             expected_response_code=200
-        )
+        ).close()
         return True
 
     # Writing Data
@@ -267,7 +267,7 @@ class InfluxDBClient(object):
                 params=params,
                 data=data,
                 expected_response_code=200
-            )
+            ).close()
 
         return True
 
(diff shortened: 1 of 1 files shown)


ReAzem:
Have you investigated why this breaks the tests?

jove1:
It works with my version of requests (2.5.3). 
Here the tests fail in the requests library:
def close():
...
return self.raw.release_conn()
AttributeError: 'NoneType' object has no attribute 'release_conn'

In docs, the close() function is mentioned in the part that deals with responses providing raw socket interface. It is possible that at some time close() only worked on responses with raw socket.

Simple fix would be s/.close()/.content/ to explicitely request all content of the response if it is ignored later. I can make the patch, however it would not be very intuitive. Even now it is quite repetitive. "Proper" fix would be to have request() with another argument that says its response is ignored and it should be closed (or have the data read) implicitely.

jove1:
Sorry. I investigated further and it seems that, the problem (new connection for each request) went away with newer version of requests.

PR #126, opened 2015-03-17
Dump directly to file

CBarraford:
This adds the ability to dump the results of a query directly to a file instead of storing in memory. This is usefully when you're pulling huge amounts of data from influxdb (ie several GBs). An example use case is a backup script that backs up influx data directly to disk.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -138,7 +138,7 @@ class InfluxDBClient(object):
         self._username = username
         self._password = password
 
-    def request(self, url, method='GET', params=None, data=None,
+    def request(self, url, method='GET', params=None, data=None, stream=False,
                 expected_response_code=200):
         """
         Make a http request to API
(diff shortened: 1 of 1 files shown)


CBarraford:
on second thought, maybe this shouldn't be apart of the `query` method. Maybe it should be its own thing like `query_to_disk`?

ReAzem:
This is a use case I did not think about. Thank you for your contribution.

>Maybe it should be its own thing like `query_to_disk`?

Yes, I think this should be a separate method.

Make sure this is usable from the outside! You have only modified the `_query` method, which is private.

Also, tests!

CBarraford:
next step is to write tests, will get to that soon.

gst:
@CBarraford 
if you check current master version and even more this branch/PR : https://github.com/influxdb/influxdb-python/pull/136,

I'd propose you add your test within client_test_with_server.py:CommonTests (class unittest.TestCase) (which is in the PR above)..

PR #130, opened 2015-03-23
add custom timestamp to SeriesHelper

areski:


ReAzem:
I don't think that this is the most intuitive way to do this. Why not just always accept the timestamp parameter when creating new points?

A simple kwargs pop could be done at [helper.py#L105](https://github.com/areski/influxdb-python/blob/fix-timeseries-timestamp/influxdb/helper.py#L105).

areski:
if you do so there then the 'cls._datapoints' will not be set correctly, this will fail:
cls._datapoints[cls._series_name.format(**kw)].append(cls._type(**kw))

ReAzem:
Sorry for the long delay :(

 I have been busy with the ResultSet. I'll get to this soon!

areski:
no worries ;)

gst:
@areski 
there is a minor flake8 issue : 

influxdb/helper.py:147:80: E501 line too long (81 > 79 characters)

are you still around there to complete entirely this PR ? :)

areski:
@gst here a fix for the flake8 issue

Can you be more specific about what you mean by completing *entirely* the PR?

gst:
> Can you be more specific about what you mean by completing entirely the PR?

nothing more, or yes if not already done : if you can rebase your branch on top of master and update the pr ?
also @aviau will have to look on this as I don't know how it's still needed ..

areski:
Let's see if @aviau wants this in, no point we work on this if it's not needed ;)

gst:
right ;)
let's wait & see..

aviau:
Can we have a tests? I'll merge after.

aviau:
@areski Do you want to finish this PR, or should we take over? 

Todo:
 - Add tests
 - Change ```timestamp``` to ```time```
 - Add a time to each point when time is not specified.

The reason we should add a time when it is not specified is explained by #264. Inserting several points with no time will result in only the last being taken.

matt-snider:
Hi is this PR still going to be developed and merged? This feature would be very useful to us so that we can use a high batch size, while still retaining good time resolution (and not losing any points).

aviau:
@matt-snider Would you like to take over?

matt-snider:
Sorry for the late response. 

I will take a look at this and open up another PR when I get some time.

aviau:
I'll close this as @matt-snider's PR was merged

PR #132, opened 2015-03-24
Fixed docs badge link

timmartin19:
It now appropriately goes to the actual documentation which most people would expect.


ReAzem:
Merged! Thank you for contributing :)

PR #133, opened 2015-03-25
Pythonic changes

gst:
Some/few pythonic (mostly obvious) changes.

The greater is the change to dataframe_client, I split it up in 2 files for cleaner separation ;  `_dataframe_client.py` is now in fact the "real"/usable one : it completely assumes that pandas is available. while `dataframe_client.py` simply wraps it but providing a fallback (as previously) DataFrameClient class which does the same thing (as previously) in case pandas isn't available/usable.


ReAzem:
Thank you for contributing :)

gst:
damn, 
```
+                                for cur_col, field in enumerate(value):
+                                    item[row['columns'][cur_col]] = field
+                                    cur_col += 1
```

the `cur_col += 1` is useless, I'm removing it also..

gst:
erf, you already merged, ok no prob doing another pr ;)

ReAzem:
hehe, indeed it is!

PR #140, opened 2015-03-29
Use a Session per InfluxDBClient instance

mattrobenolt:
It's better to avoid the global `session` instance and create a new one per instance of `InfluxDBClient`

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -14,8 +14,6 @@ try:
 except NameError:
     xrange = range
 
-session = requests.Session()
-
 
 class InfluxDBClientError(Exception):
     """Raised when an error occurs in the request"""
@@ -85,6 +83,7 @@ class InfluxDBClient(object):
 
         self.use_udp = use_udp
         self.udp_port = udp_port
+        self._session = requests.Session()
         if use_udp:
             self.udp_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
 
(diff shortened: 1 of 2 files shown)


gst:
+1

influxdb/client.py
```diff
@@ -84,6 +82,8 @@ def __init__(self,
         self.udp_port = udp_port
         if use_udp:
             self.udp_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        else:
```
gst: For the tests to succeed:
remove the "else:" part and keep the `self._session = ..` always.

ChristopherRabotin:
Does this lead to any actual performance improvement?

Requests' `Session`s helps in clarity and conciseness when dealing with auth or interacting with APIs which are not fully RESTful. However, my understanding is that InfluxDB is fully RESTful (requiring username and password for each protected request): hence, my understanding is that using requests' `Session`s would not lead to any substantial benefit, neither in clarity nor in performance.

mattrobenolt:
The biggest gain from a `Session` is keepalive. It'll maintain a keepalive with the server, which is a substantial gain.

ChristopherRabotin:
You're right, I did not know that (despite it being in the docs).

influxdb/client.py
mattrobenolt: lol I like how the tests use this when publishing a UDP metric does not. ;)

I'll fix this up.

ReAzem:
@mattrobenolt Do you intend to fix the tests?

mattrobenolt:
@ReAzem Fixed. :)

ReAzem:
Thank you for contributing :)

PR #141, opened 2015-03-30
Add a "simple" test case to verify the mean() good behavior

gst:
Quite simple test verifying the good behavior of the mean() function.

### DIFF
--- a/tests/influxdb/client_test_with_server.py
+++ b/tests/influxdb/client_test_with_server.py
@@ -10,10 +10,11 @@ This basically duplicates what's in client_test.py
 
 """
 
-from __future__ import print_function
+from __future__ import print_function, division
 
 import datetime
 import distutils.spawn
+from functools import partial
 import os
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


ReAzem:
The point of this library is not to test InfluxDB. What are you trying to do here?

gst:
yeaaah sure ;)

that's for the same reason than the others tests with a server instance running:  it ease to detect change of format response from the server..  

and as far as it's possible to use a real server instance then I find it far better than using mock of requests package.. 

no more, no less ;)

gst:
> The point of this library is not to test InfluxDB

the point is of the library is to offer a python client library to query an influxdb server.

So I find that actually testing the good behavior of this client library *when* used against a real server is the best thing to actually test the client itself. don't you think ?

gst:
I will rebase it once https://github.com/influxdb/influxdb-python/pull/142 as been merged. this one also needs it because of last server update..

ReAzem:
Using a function does not affect the response format,  so you are not testing anything meaningful here. The library does not do anything different because you have used ``mean`` in your query. 

This test would make sense in the InfluxDB repository if you wanted to test that ``30 +60+30 / 3 = 40``

gst:
> Using a function does not affect the response format

hmm, it might be, but that's not necessarily (always) the case.. The server could use different formatting function(s) depending on the query and what it contains..

but you're right that this kind of tests could (or should) also be part of the server test itself..

ReAzem:
>this kind of tests could (or should) *also* be part of the server test itself..

For the reasons stated above, I think that it should only be part of the server So I will not merge this.

Issue #143, opened 2015-03-30
SELECT with GROUP BY a tag return only 1 row


gst:
If you have a serie with distinct tags value then "SELECT * FROM serie_name GROUP BY a_tag_name" only returns one result : while it should returns one for each distinct serie_name:a_tag_name couple..

it's (I think) all is within the `format_query_response()` function which I have to analyse a bit to see how to handle correctly that..

ReAzem:
> it's (I think) all is within the `format_query_response()` function which I have to analyse a bit to see how to handle correctly that..

This is something I have not tested yet, so you must be right! A test case would be very welcome :)

gst:
> This is something I have not tested yet, so you must be right! A test case would be very welcome :)

I've already begun ;)

gst:
Fixer un PR 144

PR #144, opened 2015-03-30
Issue 143

gst:
Fix issue #143 

Adapted `format_query_response()` to also handle tags returned.

I'd like some advices on the format I use to distinguish same serie with different tags.
Basically : if there are some tag(s) returned by a query then store them in the result by providing a unique entry for each such unique serie + tag(s) value(s) with `name = (row['name'], tuple(tags.values()))`

Now after thinking about it twice I think in fact I should also include the tags.keys() somewhere.. also more than probably in the `name` attribute of each result dict with `name = (row['name'], tuple(tags.items()))`   (*items* simply instead of *values* so) ; so that the caller can know which tag value correspond to which tag name..  otherwise if the groupby is done on multiple tags so then the caller can't know for sure the order of the tags as returned in the answer.

wdyt ?

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -2,6 +2,7 @@
 """
 Python client for InfluxDB
 """
+from collections import OrderedDict
 import json
 import socket
 import requests
@@ -28,6 +29,7 @@ class InfluxDBClientError(Exception):
 
 
 class InfluxDBClient(object):
+
     """
     The ``InfluxDBClient`` object holds information necessary to connect
     to InfluxDB. Requests can be made to InfluxDB directly through the client.
@@ -67,7 +69,8 @@ class InfluxDBClient(object):
                  verify_ssl=False,
                  timeout=None,
                  use_udp=False,
-                 udp_port=4444):
+                 udp_port=4444,
+                 ):
         """
         Construct a new InfluxDBClient object.
         """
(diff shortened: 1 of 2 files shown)


ReAzem:
I think that the series identifier should be a tuple with a name and a dict.

for example: ``("cpu", {"host_name": "example.com"})``

gst:
> I think that the series identifier should be a tuple with a name and a dict.
> for example: ("cpu", {"host_name": "example.com"})

yep, that's what I also get in my second thought presented here above.

Problem is that the `name` attribute of the main dict result (from the `format_query_response` function) is a key (in the dict result so), so we can't use a dict as one of its component (not hashable).. so my `tags.items()` good second choice I think..

ReAzem:
> so we can't use a dict as one of its component (not hashable)

You are right! :(

> so my tags.items() good second choice I think..

Absolutely.

gst:
will travis make it succeed on the first pass ?

ReAzem:
> will travis make it succeed on the first pass ?
What?

gst:
> What?

I mean  : "is travis going to succeed the tests without any temporary error" :p

ReAzem:
Looks like it failed for a valid reason. The order of tags is problematic. Looks like we will have to sort them.

gst:
fixed in the test.

I'm not in favor of sorting things "undesirably" in the library itself. If one wants the sorted serie names from the library it should re-use the same "trick".. after its query() call. 

So I could eventually put the function `sorted_series_tags` in the library itself. (with a better name eventually ? if you have one..)  or not at all. 
what do you think ?

ReAzem:
We are not sorting the series but just the tags,  so I don't see why this is underisable.

gst:
Hmm, makes sense..

influxdb/client.py
```diff
@@ -67,7 +69,9 @@ def __init__(self,
                  verify_ssl=False,
                  timeout=None,
                  use_udp=False,
-                 udp_port=4444):
+                 udp_port=4444,
+                 keep_json_response_order=True,
```
ReAzem: Add it to the docstring please.

gst:
Seems all good..

ReAzem:
Seems good to me too!

PR #146, opened 2015-03-31
Pythonic: `x in dict.keys()` => `x in dict`

gst:


ReAzem:
Is there a downsite to not using ``.keys()``?

gst:
on python2 this uselessly call the keys() function, which uselessly creates a list of the keys and then return it. and then the "in" operator is finally applied to it.
(on python3 this uselessly call the keys() function, which uselessly returns an iterator over the dict keys).

otherwise it's simply the recommended/pythonic syntax to check for a key in a dict ;)

ReAzem:
Alright, thanks!

gst:
> Is there a downsite to not using .keys()?

last but not least : 

doing "x in dict.keys()" (in py2 as well as py3) is a linear search : O(n) 
while doing "x in dict" uses the dict internal hashing : O(1)

PR #148, opened 2015-04-02
Add a client `InfluxDBClusterClient` to handle a cluster of InfluxDB servers

cannium:


cannium:
I'll try to figure out how to properly test the class. But before that I must fix my InfluxDB cluster where I seems to hit bugs.

gst:
> What would be incredible is a(some) test case(s), against a real cluster, where you "kill" (in a way or another) some of the cluster nodes and you assert that queries are always executed (and get the correct answer) despite some of the nodes being killed/shutdown..

I have to agree/say that it's not necessarily easy to get that..  Doing it with mocks could be quite more easy to get working.. to be seen..

cannium:
Modified code pushed, tests and docs comming

cannium:
Seems the failure isn't caused by this patch set, how can I trigger a retest?

ReAzem:
You can't. I just did!

ReAzem:
I intend to review this soon, sorry for the delay!

tests/influxdb/client_test.py
```diff
@@ -534,3 +535,97 @@ def connection_error(self, *args, **kwargs):
...
+
+    def test_all_fail(self):
+        cluster = InfluxDBClusterClient(hosts=self.hosts,
+                                        database='database',
+                                        shuffle=True,
+                                        client_base_class=FakeClient)
+        try:
+            cluster.query('Fail')
+        except InfluxDBServerError:
+            pass
```
gst: You could/should also "assert the whole try/except block" with : 
```
with self.assertRaises(InfluxDBServerError) as ctx:
    cluster.query('Fail')
```
?

cannium: done

tests/influxdb/client_test.py
```diff
@@ -534,3 +535,97 @@ def connection_error(self, *args, **kwargs):
...
+
+    def test_recovery(self):
+        cluster = InfluxDBClusterClient(hosts=self.hosts,
+                                        database='database',
+                                        shuffle=True,
+                                        client_base_class=FakeClient)
+        try:
+            cluster.query('Fail')
+        except InfluxDBServerError:
+            pass
```
cannium: Here [as well](https://github.com/influxdb/influxdb-python/pull/148/files#r28265757)

ReAzem:
Looks great, but my one issue with this is that I wonder how other clients will be able to use it.

For example the DataFrameClient and the SeriesHelper.

Would there be an acceptable way of including this in the InfluxDBClient?

cannium:
By setting `client_base_class=xxx` I suppose?

ReAzem:
Oh, I had overlooked! Makes sense!

However, other clients may need other things in the init.

Also, maybe we should use from_DSN?

cannium:
Seems current `from_DSN` code in current code can not handle connection strings like `influxdb://username:password@host1:port1,username:password@host2:port2,username:password@host3:port3/database_name`, as in [some other db](http://docs.mongodb.org/manual/reference/connection-string/#standard-connection-string-format). 
What I'm thinking about is to create a function like `create_new_client`, if the connection string given includes a single server, returns an instance of `InfluxDBClient`, otherwise return an instance of `InfluxDBClusterClient`. What do you think?

ReAzem:
@cannium Sorry for the delay.

It looks like this is ready to be merged! Can I proceed? :)

cannium:
Sure, go ahead

ReAzem:
Thank you for your contribution!

PR #150, opened 2015-04-03
Remove dict comprehension for older python

cannium:
dict comprehension is introduced in python 2.7 and doesn't work on 2.6 or older


ReAzem:
Is this the only thing that is not compatible 2.6? I have never tested the lib with python2.6. Is it really worth supporting?

cannium:
I found it when I tried to pack an RPM for centos6.5 and python2.6 complained about it doesn't know how to handle the syntax. So I could safely assume that's the only syntax not compatible with 2.6. For other parts as far as I've tested, runs well with 2.6.

ReAzem:
Alright then!  Thank your for contributing :)

ReAzem:
Merged Remove dict comprehension for older python #150

PR #151, opened 2015-04-07
Resultset try42 [WIP]

gst:
I create a PR with this unfinished work because there are certain things we must decide..

1) Returning a ResultSet from InfluxDbClient.query() is great.

2) Enabling to get the full list of resulting *series* from a resultset by doing `list(resultset)` is also great. But we have to clearly define the format of what's returned in the list actually.. 

3) Enabling to get the full list (eventually filtered by serie name and/or tags values) of resulting *points* from a resultset using `resultset[key]` is also great. Here the `Point`s are more well defined within influxdb.point now.

to be continued..


influxdb/resultset.py
```diff
@@ -91,14 +91,14 @@ def __len__(self):
     def keys(self):
         keys = []
         for serie in self._get_series():
-            keys.append((serie['name'], serie['tags']))
+            keys.append((serie.get('name', None), serie.get('tags', None)))
```
ChristopherRabotin: In which cases would `name` not be present? Same question for `tags`? Is it spec'd that no tags returns an empty list from the InfluxDB API: in which case that missing key should raise an exception instead of returning confusing data `(None, None)`, wouldn't you think?

gst: "system" queries do not return a "name" , and "normal" queries (no "group by tag_name1(, tag_name2, ..)") doesn't return tags either.

NB: it's quite work in progress here ;)

ReAzem:
I'll open a new PR with all the ResultSet code.

ReAzem:
see #154

PR #154, opened 2015-04-08
Resultset

ReAzem:


gst:
Not that I'm against that you merge your own PR , but for this one there could/should have been **at least** a big squash of the vast majority of the *temporary/tests* commits.. 

and some few others things too that I'd have liked to comment on the final result..  it's never too late, in resultset.py : 

```
    def __iter__(self):
        """ Iterating a ResultSet will yield one dict instance per serie result.
        """
        for key in self.keys():
            yield list(self.__getitem__(key))
```
1) docstring: it's not (anymore) dict instances which are yielded, but lists (of "points")
2) self.__getitem__(key) <=> self[key]  # the later being the preferred / pythonic way..

ReAzem:
> Not that I'm against that you merge your own PR , but for this one there could/should have been at least a big squash of the vast majority of the temporary/tests commits.. 

We both worked on the PR and I didn't want to remove your authorship. The good thing to do would have been to squash it all under one single commit :( But you're right. I still could have squashed commits in-between.

I'll wait for ResultSet to be stable before I make a new release, so its best to work on master.

gst:
> We both worked on the PR and I didn't want to remove your authorship. 

for such temporary crappy test commits it's ok ;) 

> The good thing to do would have been to squash it all under one single commit :( 

ultimately yep.
we'll see as you say.

PR #158, opened 2015-04-13
Revert "Revert "Sync tests against last master update""

gst:
This reverts commit d3bb67d4793435274a6096169835715179cbe8d4.

This commit was "good" : 
* it slightly simplified the server setup & launch 
* and it also fix the webui port (was missing) (And more important this makes it possible to test against the last server RC (rc23)).


ReAzem:
Waitint for Travis

gst:
> and it also fix the webui port (was missing) (And more important this makes it possible to test against the last server RC (rc23)).

though, they have added some info in the response in case of db not previously existing

AssertionError: '{"results":[{"error":"database not found"}]}' != u'{"results":[{"error":"database not found: db (/home/gstarck/go/src/github.com/influxdb/influxdb/server.go:1010)"}]}'

.. grrr.. will have to adapt the assert..

gst:
> AssertionError: '{"results":[{"error":"database not found"}]}' != u'{"results":[{"error":"database not found: db (/home/gstarck/go/src/github.com/influxdb/influxdb/server.go:1010)"}]}'

> .. grrr.. will have to adapt the assert..

we will see when we'll update the server-RC we use ;)

ReAzem:
@gst You can go ahead and update it in the .travis.yml

gst:
rebased on master.

But they may change again that response content before final release : 
https://groups.google.com/forum/#!topic/influxdb/YO0bKLV9Zfs

so we can just wait that they do that ;)

ReAzem:
@gst You have python 3.2 failing. Should be an easy-fix.

PR #160, opened 2015-04-21
Fix for get_list_users method in InfluxDBClient

georgijd:
Hi guys,

Noticed a small bug today with the get_list_users method returning list of lists instead of just a list and this is the proposed fix with included test cases.

Cheers!


ReAzem:
Hello Georgi,

Looks good to me! Thank you for contributing!

ReAzem:
I have just released 2.0.2 and pushed it to PyPI.

PR #162, opened 2015-04-22
Add user management operations

georgijd:
Hi there,

I added a few methods for user management - create_user, drop_user and set_user_password. Tests are included as well, except for the set_user_password, because I noticed that the actual influxdb query (SET PASSWORD FOR user = 'password') doesn't seem to do much (it appears to me that even after the password is changed one can still query influxdb with the old credentials). When I have more time I'll investigate if that is the case and add tests for that method as well.

Cheers!


ReAzem:
@georgijd 

Very good PR, thanks!

Issue #108, opened 2015-02-28
DataFrameClient with 0.9.0


aviau:
@timtroendle Would you be willing to test the DataframeClient with InfluxDB 0.9.0?

There is influxdb 0.9.0 support in the 0.9.0_support branch.

timtroendle:
I don't have an environment to test at the moment. Maybe @tzonghao could support?

gst:
actually the DataframeClient tests completely fail because there are changes required to handle the changes in the format of the responses  we get from influxdb 0.9.

gst:
well, not completely but nearly : 

```
Ran 11 tests in 0.064s 
FAILED (errors=8)
```

aviau:
This was released

PR #163, opened 2015-04-23
DataFrameClient for InfluxDB 0.9

tzonghao:
#108 

* `get_list_series` returns a mapping from measurement names to DataFrame representing tags for the named measurement.
* `query` returns a mapping from (measurement, (tag (k, v) pairs), to DataFrame representing time series for the measurement (qualified by tags) in question. Query without tags should simply use measurement as keys.
* `write_points` converts a mapping of DataFrame objects as `points`, then hand them over to super's `write_points`.


ReAzem:
Hi,

Thank you for taking on this issue :)

Have you looked at ``client_test_with_server.py``? I would like some DaframeClient tests there.

You are using the ``u'string'`` syntax, which does not work in python 3.2.

Other than that, this looks great!

//cc @timtroendle

tzonghao:
@ReAzem In case you haven't reviewed it, a few basic tests were added in `client_test_with_server.py` in this pull request as suggested. The 3.2 syntax problem is also fixed. I've also changed the key data structure for data dictionary's keys, for both querying and submitting, where strings can be used as keys if no tags are used. Otherwise, the keys can be a tuple of (measurement, sorted tuple of tag key-value pairs).

ReAzem:
@tzonghao Hello,

Than you for the great work. Why are you not using the ResultSet Object? You should not have to use the raw property.

tzonghao:
@ReAzem You're right. Using ResultSet is cleaner. It's converted in the last commit.

ReAzem:
@tzonghao Can you change the write_points function to accept three arguments instead?

- measurment name
- tags (optional)
- dataframe

I think it would be more intuitive than a tuple of key-value tuples.

tzonghao:
@ReAzem Sure, I can make that change. It also implicates that `write_points` will accept one DataFrame at a time, instead of multiple DataFrames.

ReAzem:
@tzonghao  I think it would be worth it, do you agree?

tzonghao:
I agree. It's a good decision.

tzonghao:
@ReAzem It looks like an InfluxDB error in the last Travis run that failed the py34 test. Can you trigger a rerun?

examples/tutorial_pandas.py
```diff
@@ -22,6 +22,9 @@ def main(host='localhost', port=8086):
     print("Write DataFrame")
     client.write_points({'demo': df})
 
+    print("Write DataFrame with Tags")
+    client.write_points({('demo', (('k1', 'v1'), ('k2', 'v2'))): df})
```
ReAzem: I don't think this works anymore

tzonghao: good catch

ReAzem:
Great work! I will be merging this for now. I'm sure we can keep on improving it after feedback from other users.

Issue #164, opened 2015-04-23
JSON response decoding error with InfluxDB 0.8


mnuessler:
I tried to use the InfluxDBClient to query InfluxDB v0.8.8 and get the following error:

```
Traceback (most recent call last):
  File "./test.py", line 4, in <module>
    result = client.query('select * from /.*/ limit 1', chunked=True)
  File "/home/matt/projects/influxdb-python-test/venv/local/lib/python2.7/site-packages/influxdb/influxdb08/client.py", line 419, in query
    chunked=chunked)
  File "/home/matt/projects/influxdb-python-test/venv/local/lib/python2.7/site-packages/influxdb/influxdb08/client.py", line 451, in _query
    return list(chunked_json.loads(response.content.decode()))
UnicodeDecodeError: 'ascii' codec can't decode byte 0x8b in position 1: ordinal not in range(128)
```

The response returned by the InfluxDB is chunked and also compressed with gzip. It seems that the client tries to parse the gzipped data as JSON. 

I am using the client for InfluxDB 0.8 as you can see below.
```python
from influxdb.influxdb08 import InfluxDBClient                                                                                                                                                                
                                                                                                                                                                                                              
client = InfluxDBClient('localhost', 8086, 'myusername', 'mypassword', 'mydatabase')                                                                                                                                
result = client.query('select * from /.*/ limit 1')                                                                                                                                             
                                                                                                                                                                                                              
print("Result: {0}".format(result)) 
```

Relevant HTTP headers as sent in the request:
```
Accept: text/plain
Content-Type: application/json
Accept-Encoding: gzip, deflate
```
Relevant HTTP headers as received in the response:
```
Content-Encoding: gzip
Content-Type: application/json
X-Influxdb-Version: InfluxDB v0.8.8 (git: afde71e) (leveldb: 1.15)
Transfer-Encoding: chunked
```
Python package versions:
```
$ pip list
influxdb (2.0.2)
requests (2.6.1)
```

Shouldn't the Python InfluxDB (or the requests library) decompress the response before parsing? Or should the InfluxDB API send a response with uncompressed data?

georgijd:
Hi there,

I think I found the problem. It's not that the client is trying to parse the gzipped as JSON, but most probably you are storing strings with unicode characters as values and the decode() method is having a hard time decoding those unicode characters.

georgijd:
Hi,

[...]

PR #166, opened 2015-04-23
Fix issue #164

georgijd:
This commit should fix Issue #164 . As usual, a test is included.

### DIFF
--- a/influxdb/influxdb08/client.py
+++ b/influxdb/influxdb08/client.py
@@ -448,7 +448,13 @@ class InfluxDBClient(object):
         )
 
         if chunked:
-            return list(chunked_json.loads(response.content.decode()))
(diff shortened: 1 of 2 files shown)


georgijd:
Hi,

Although this doesn't actually fix #164, since it's not a influxdb-python bug, it still fixes a bug that appears when the response content contains unicode characters. I'm not sure why the last build failed, may be you can take a look at it. Codewise everything looks good to me.

Cheers!

ReAzem:
Hello,

Thank you for the PR. It all looks good to me too.

I think @gst it troubleshooting the build issue right now.

ReAzem:
The failure has nothing to do with your PR, so I will merge this right now!

Thank you for your contribution :)!

PR #168, opened 2015-04-24
clean: not anymore used/necessary.

gst:
was preventing influxdb-python to be importable in python2.6 .. (because of OrderedDict)


ReAzem:
Yep, this was dead code! Thanks!

PR #178, opened 2015-05-03
Allow mixed case in DSN (closes #176)

### LINKED ISSUE #176: from_DSN won't work with upper case passwords
Hello,

If the password given in the dsn string to InfluxDBClient.from_DSN contains some upper case letters, then subsequent queries on the client will fail with:
```
  File "/usr/local/lib/python2.7/site-packages/influxdb/influxdb08/client.py", line 419, in query
    chunked=chunked)
  File "/usr/local/lib/python2.7/site-packages/influxdb/influxdb08/client.py", line 447, in _query
    expected_response_code=200
  File "/usr/local/lib/python2.7/site-packages/influxdb/influxdb08/client.py", line 255, in request
    raise InfluxDBClientError(response.content, response.status_code)
influxdb.influxdb08.client.InfluxDBClientError: 401: Invalid username/password
```

Good practice for passwords is to mix lower and upper cases.
I guess other params like username and database name in the dsn could also have this problem if they are case sensitive as well. I did not test them.

For the moment it can be solved by giving the password as an extra parameter to from_DSN. But it could confuse other users and take them a while to understand.

Cheers,
Axel

ReAzem:
@cannium What do you think?

That would close #176

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -144,7 +144,6 @@ localhost:8086/databasename', timeout=5, udp_port=159)
                   additional udp_port parameter (cf. examples).
         :raise ValueError: if the provided DSN has any unexpected value.
         """
-        dsn = dsn.lower()
 
         init_args = {}
         conn_params = urlparse(dsn)
@@ -559,7 +558,6 @@ class InfluxDBClusterClient(object):
         :param client_base_class: In order to support different clients,
                                   default to InfluxDBClient
         """
-        dsn = dsn.lower()
         conn_params = urlparse(dsn)
         netlocs = conn_params.netloc.split(',')
         cluster_client = InfluxDBClusterClient(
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
I think the reason why I converted this to lower case is to make sure that the protocol is correctly identified even if it's given in uppercase. Maybe would it be worth doing the protocol check with the lowercase version of the DSN.

ReAzem:
Oh. I pinged the wrong guy.  My bad :) 

You are right,  it would be a good idea convert the case to check for the protocol.

I'll give a try tomorrow.

cannium:
@ReAzem No problem.
The scheme part(`https+influxdb`) should be converted to lower case to check, and other parts(like username, password) are case sensitive, in my opinion.

ReAzem:
Looks like it already worked. ``urlparse`` does the work for us.

ReAzem:
@ChristopherRabotin Any feedback? Good to merge?

ChristopherRabotin:
@ReAzem looks good, nothing to add. Thanks for asking!

ReAzem:
Thank you :)!

PR #190, opened 2015-06-01
Remove dsn.lower() from influxdb08.InfluxDBClient.from_DSN()

kipe:
InfluxDBClient.from_DSN() now behaves correctly and allows (for example) case-sensitive password.

### DIFF
--- a/tests/influxdb/client_test.py
+++ b/tests/influxdb/client_test.py
@@ -29,6 +29,9 @@ import mock
 from influxdb import InfluxDBClient, InfluxDBClusterClient
 from influxdb.client import InfluxDBServerError
 
+DSN_STRING = 'influxdb://uSr:pWd@host:1886/db'
+CLUSTER_DSN_STRING = 'influxdb://uSr:pWd@host1:8086,uSr:pWd@host2:8086/db'
+
 
 def _build_response_object(status_code=200, content=""):
     resp = requests.Response()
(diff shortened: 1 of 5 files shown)


ChristopherRabotin:
Thanks for spotting that. Could you add to the `influxdb08` package the similar tests that we added to the main package? https://github.com/influxdb/influxdb-python/blob/master/tests/influxdb/client_test.py#L810

Thanks.

kipe:
In my opinion these kinds of problems are easiest to capture by using the existing tests, but changing the case-sensitive fields (`username` and `password`, others?) to include values which use uppercase letters.

So for example [tests/influxdb/influxdb08/client_test.py#L102](https://github.com/influxdb/influxdb-python/blob/master/tests/influxdb/influxdb08/client_test.py#L102) would change from
`cli = InfluxDBClient.from_DSN('influxdb://usr:pwd@host:1886/db')`
to
`cli = InfluxDBClient.from_DSN('influxdb://uSr:pWd@host:1886/db')`

Also, for easier change (and avoiding typos) the "base url" should be one variable, instead of rewriting it all the time (e.g. `DSN_STRING = 'influxdb://uSr:pWd@host:1886/db'`). This way the same "base url" could be used in all the cases, just by using `'udp+' + DSN_STRING` for UDP-test etc.

I can try and do this today or tomorrow, but if this is the way to go, I'd rather create another pull request, as it affects multiple tests (I'd also do the changes to 0.9 at the same time).

kipe:
Just added some refactoring of the tests, which also modify the tests to include the situation this pull request fixes.

tests/influxdb/client_test.py
```diff
@@ -29,6 +29,9 @@
 from influxdb import InfluxDBClient, InfluxDBClusterClient
 from influxdb.client import InfluxDBServerError
 
+DSN_STRING = 'influxdb://uSr:pWd@host:1886/db'
+CLUSTER_DSN_STRING = 'influxdb://uSr:pWd@host1:8086,uSr:pWd@host2:8086/db'
```
aviau: Can you declare this in a set_up instead? I prefer not to have class-level attributes

kipe:
Good point, changed as requested :)

aviau:
Thank you for your contributon @kipe !

@ChristopherRabotin  Is that good to you?

ChristopherRabotin:
Yes, that looks good. Thanks for asking.

PR #194, opened 2015-06-14
Refactored ResultSet

### LINKED ISSUE #192: Multiple query request to same  series creates a bad ResultSet
When sending multiple queries in the same request it can result in a bad ResultSet when querying from the same series.

For instance if I run these two queries in the same request:

<code>
"select * from foo where time > now() - 20; select * from foo where time < now() - 40"
</code>

I believe it is because the ResultSet uses the series name as the key meaning that both these query results can't be in the ResultSet.

aviau:
Improvements on the ResultSet API to make it more intuitive.

* Replaced ``__getitem__`` by ``get_points()``
* Added warnings on ``__getitem__``
* Raise errors when they are present in ResultSets
* closes #192 - Return several ResultSets when apropriate
* Re-enable tests that were broken due to the fact that HTTP errors are no longer returned

### DIFF
--- a/docs/source/resultset.rst
+++ b/docs/source/resultset.rst
@@ -7,33 +7,39 @@ Query response object: ResultSet
 
 Using the ``InfluxDBClient.query()`` function will return a ``ResultSet`` Object.
 
-A ResultSet behaves like a dict. Its keys are series and values are points. However, it is a little bit smarter than a regular dict. Its ``__getitem__`` method can be used to query the ResultSet in several ways.
+A ResultSet can be browsed in several ways. Its ``get_points`` method can be used to retrieve points generators that filter either by measurement, tags, or both.
 
-Filtering by serie name
------------------------
+Getting all points
+------------------
 
-Using ``rs['cpu']`` will return a generator for all the points that are in a serie named ``cpu``, no matter the tags.
+Using ``rs.get_points()`` will return a generator for all the points in the ResultSet.
+
+
+Filtering by measurement
+------------------------
+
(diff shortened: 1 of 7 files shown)


aviau:
@ChristopherRabotin, @adventures91,  @kipe, @weiyoung-zhou @georgijd 

Any feedback?

kipe:
I do like the changes in general, makes the code more pythonic.
Especially the filtering by tags seems more logical by using this kind of structure.

It would be more logical to me, if one query would return a single ResultSet and multiple queries a list of ResultSets. 

On the other hand, returning a list always would be constant, but does raise the question on how often is this feature used. And even if it is, the programmer should know he's running multiple queries and should expect multiple results -> knows to expect a list of ResultSets instead of a single ResultSet.

aviau:
> It would be more logical to me, if one query would return a single ResultSet and multiple queries a list of ResultSets. 

That's exactly what it does in my PR.

>  but does raise the question on how often is this feature used.

Indeed. Most users will have to get the first element of the list all the time. It would be quite annoying.

georgijd:
I second @kipe's opinion on the filtering :) That way the ResultSet object becomes more flexible and we can easily add more filter parameters in the future. As for the multiple queries result type, what makes most sense to me is that a single query should return a single `ResultSet` object and multiple queries should return a list of `ResultSet` objects. From developer's perspective, it's really frustrating to always use expressions like `rs[0]` to get the data you need when most of the time you execute a single query. It pollutes the code and it's prone to errors.

aviau:
Thank you for the feedback.

PR #196, opened 2015-06-15
Allows python `NaN` values to be written as `null` in the JSON data

svernede:
JSON dump modified to allows python `NaN` values to be written as `null` in the JSON data

see https://github.com/influxdb/influxdb-python/issues/195

Adds a new dependency to module   `simplejson` but the same effect is difficult to achieve with base module `json`
see http://stackoverflow.com/questions/28639953/python-nan-json-encoder

### DIFF
--- a/influxdb/influxdb08/client.py
+++ b/influxdb/influxdb08/client.py
@@ -7,6 +7,7 @@ import socket
 import requests
 import requests.exceptions
 import warnings
+import simplejson
 from sys import version_info
 
 from influxdb import chunked_json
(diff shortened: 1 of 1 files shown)


aviau:
I can see the improvement. However, I don't really like adding simplejson as a dependency. Are there any other solutions?

svernede:
Yes, I understand adding a new dependency is not desirable.
I could not find a way to achieve the result simply with the base `json` module. 
Any suggestions are welcomed

georgijd:
Hi guys,

May I suggest an alternative solution to this problem? If the dataframe value is a list of floats, all NaN values in that list will be replaced with None thus making the numpy array object JSON compliant. The nice thing about this is that all of it happens before the serialization takes place. See one possible implementation here: 0feadd926f. If you're happy with it, I'll make a pull request.

Cheers!

svernede:
Looks good to me.
Thanks you Georgi for your kind help

aviau:
Closed in favor of #204

PR #202, opened 2015-06-16
New query option: raise_errors (Closes #201)

### LINKED ISSUE #201: Multiple queries with a series that doesnt exist shouldn't throw an exception
When querying multiple series if one of the series does not exist it throws an exception and you cant get any results.

Some series are based off of continuous queries that may not have run yet. So I cant really know if the series exist yet without multiple queries to the database to first check if a given series exists.

I am not sure of the best solution to this but I don't think the client should throw an exception for this specific error case.

aviau:


aviau:
Waiting for Travis.

@adventures91 Any feedback?

adventures91:
Thanks for the quick update on that and it looks good to me.

PR #206, opened 2015-06-19
Update resultset.rst

sarvesh-ranjan:
Was getting error: ResultSet has no attribute get_points()
Also getting syntax error in line 25.
Edited line 25 does the same manipulation to return the list of all the points.


aviau:
What version of influxdb-python are you using? 

Those are not equivalent at all.

Issue #191, opened 2015-06-11
Add support for InfluxDB line protocol write API


victorhooi:
InfluxDB recently introduced a new line protocol write API, for more efficient bulk entry of points:

https://github.com/influxdb/influxdb/pull/2696

It would be awesome if support for this could be added in the Python client!

trehn:
I'd also like to point out that UDP writes are currently broken since JSON over UDP was dropped in 0.9.0 already (in favor of this new line protocol).

Tempted to take a stab at this if @aviau isn't working on it already.

aviau:
@trehn I'm not yet working on it.  Your contribution would be much appreciated!  Go ahead and I'll review :)

trehn:
I have an initial implementation here: https://github.com/trehn/influxdb-python/commits/feature/line-protocol

* So far there is no reason to break backwards compatibility for the Python API.
* I introduced a dependency on python-dateutil in order to parse the string dates the Python API expects. This has the added advantage of allowing more date formats (dateutil will recognize a lot of different formats).
* `datetime` objects can now also be provided as timestamps (pretty much a side effect of parsing the date strings).
* If users want to avoid the performance overhead resulting from date string parsing and `datetime` handling, they can just pass in nanosecond integers.
* This is not implemented yet, but maybe we should use `warnings.warn` to warn people when they're specifying a database while using UDP. Selecting the database is now done on the server side for UDP and cannot be influenced by the client. Silently ignoring the database param might be confusing for users.
* The branch linked to above still has lots of test failures. If someone could take a look at these, it would be much appreciated, as I don't understand all of them (some might be obsolete now etc).

ynezz:
@trehn Thanks for your effort, it seems like InfluxDB is still fast moving target as I had to do following [small fixes](https://github.com/ynezz/influxdb-python/commit/8fe609352d952f31f331250d5319a3b2500af038) on top of your changes to get it working again.

aviau:
Please make sure that the line protocol is optional. Maybe you can open a PR?

trehn:
Why should the line protocol be optional? From what I read in https://github.com/influxdb/influxdb/pull/2696 it sounds like JSON is going away.

I'll merge the changes from @ynezz into my branch and try to get rid of the test failures.

aviau:
> it sounds like JSON is going away.

I didn't catch that at first. You are right then!

tima:
[...]

PR #209, opened 2015-06-26
Support for Line Protocol

trehn:
So this is what I have. THIS IS NOT COMPLETE, MORE WORK WILL BE REQUIRED.

There are a few test failures left. I haven't even tested this manually against a real InfluxDB instance.

Some of the fixed test cases should be closely reviewed (probably along with all the other stuff that is currently skipped because it's known to break with 0.9.0). New tests might also be required.

Sorry to leave this so messy, but I don't have any time left to commit to this effort so... take it or leave it. I think `line_protocol.py` is probably good to go, but the extra GET parameters like precision and retention probably need some love. 

See my posts in #191 for some additional notes.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -198,7 +198,7 @@ localhost:8086/databasename', timeout=5, udp_port=159)
         self._password = password
 
     def request(self, url, method='GET', params=None, data=None,
-                expected_response_code=200):
+                expected_response_code=200, headers=None):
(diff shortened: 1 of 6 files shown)


aviau:
Thank you for your PR :) I intend to take a look at it in the next few days.

aviau:
I have merged this and rebased in the line_protocol branch. 
see #211

PR #211, opened 2015-07-03
Line protocol

aviau:
- Impement line protocol (new default)
- Fix UDP writes
- New epoch parameter for queries

Thanks to @trehn and @ynezz 

\\\cc @ChristopherRabotin  @georgijd Any feedback? Would any of you guys want to test?


influxdb/client.py
```diff
@@ -96,7 +97,8 @@ def __init__(self,
 
         self._headers = {
             'Content-type': 'application/json',
-            'Accept': 'text/plain'}
+            'Accept': 'text/plain'
```
ChristopherRabotin: Any specific reason for this new line?

influxdb/client.py
```diff
@@ -203,17 +205,13 @@ def request(self, url, method='GET', params=None, data=None,
...
 
-        auth = {
-            'u': self._username,
-            'p': self._password
-        }
-
-        params.update(auth)
-
-        if data is not None and not isinstance(data, str):
+        if isinstance(data, dict) or isinstance(data, list):
```
ChristopherRabotin: AFAIK, `isinstance` supports a tuple as the second argument, so this line can be written as `if isinstance(data, (dict, list)):`.

aviau: Fixed! Thanks

influxdb/client.py
aviau: Not really, this was included in a ``Formating improvements`` commit.

influxdb/line_protocol.py
```diff
@@ -0,0 +1,103 @@
...
+    else:
+        return str(value)
+
+
+def _force_text(data):
+    """
+    Try to return a text aka unicode object from the given data.
+    """
+    if isinstance(data, binary_type):
+        return data.decode('utf-8', 'replace')
```
ChristopherRabotin: Maybe would it be worth noting in the documentation that the library does a replace when decoding `utf-8` since that could be a surprise for some users. Thoughts?

georgijd:
I'm probably going to have a in-depth look at this tommorow with some manual testing :). It looks really nice and clean at first reading.

influxdb/line_protocol.py
georgijd: I'd say make it optional. May be for some users it's going to be helpful if they know that the data they're feeding is not sucessfully processed.

aviau: I'll disable it for now. We can add an option later. I'm not sure this is a feature we want.

aviau:
I'm waiting for tests and I'll merge

Issue #223, opened 2015-07-31
Slow write performance compared to v2.6


tzonghao:
Run `%timeit` in IPython to measure performance of writing 6000 points in both influxdb-python v2.6 and v2.7.3:

```python
import numpy as np
from pandas import DataFrame, date_range
rng = date_range('2015-1-1', periods=6000, freq='s')
data = DataFrame({'value': np.random.randn(6000)}, index=rng)
measurement = 'perf_test'
tags = {"a": "val-a", "b": "val-b", "c": "val-c"}
%timeit -n3 -r10 dfclient.write_points(data, measurement, tags=tags, time_precision='u', batch_size=500)
```

Performance in v2.6:
```3 loops, best of 10: 390 ms per loop```

Performance in v2.7.3:
```3 loops, best of 10: 1.02 s per loop```

It is significantly slower.

**Comment by aviau** [comment]
I'm sure this is something that we can improve on.

However, we have to keep in mind that a slower performance makes sense because in the pas we had no processing to do client-side. We simply sent json to the server. Now, with the line protocol, we need to transform the points into lines.

**Comment by tzonghao** [comment]
That was my original suspicion as well. However, compared to the the overhead in `dateutil.parser`, the string concatenations seem insignificant. Here's a profile of writing 1500 data points (profile result truncated):

```
   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.374    0.374 <string>:1(<module>)
        1    0.000    0.000    0.374    0.374 _dataframe_client.py:33(write_points)
        3    0.000    0.000    0.359    0.120 client.py:329(write_points)
        3    0.000    0.000    0.359    0.120 client.py:384(_write_points)
        3    0.000    0.000    0.359    0.120 client.py:245(write)
        3    0.044    0.015    0.325    0.108 line_protocol.py:61(make_lines)
     1500    0.007    0.000    0.239    0.000 line_protocol.py:13(_convert_timestamp)
     1500    0.003    0.000    0.212    0.000 parser.py:945(parse)
     1500    0.015    0.000    0.207    0.000 parser.py:343(parse)
     1500    0.043    0.000    0.174    0.000 parser.py:445(_parse)
     1500    0.013    0.000    0.104    0.000 parser.py:173(split)
    24000    0.007    0.000    0.087    0.000 parser.py:170(next)
    24000    0.009    0.000    0.080    0.000 parser.py:163(__next__)
[...]

Issue #221, opened 2015-07-27
Nanosecond time data being sent even if time_precision set to something else


dschallis:
In 2.7.1, if I have e.g.:

    data = [{'fields': {'value': 12629},
                 'time': '2015-07-27T12:55:48.165231Z',
                 'measurement': 'foo'}]
    client.write_points(data, time_precision='u')

Then the 'precision' field is correctly set, but the timestamp is sent as a nanosecond integer value, e.g. the above is sent with:

    POST /write?db=test&precision=u
    ...skipped other headers here....
    foo value=12629 1438001748165231104\n

This leads to it being silently rejected by InfluxDB (i.e. after the above, a `SELECT * FROM foo` returns 0 results), which is probably a bug on their side too (though they don't document very well what should happen in this case).

(apologies for the many tickets by the way, I'm in the process of switching to this module from a reasonably sized code base, so starting to use this module quite heavily :) )

aviau:
> apologies for the many tickets by the way

Don't worry, your feedback is appreciated.

Actually, the time_precision parameter is not meant to convert your time precision. So I guess that's a feature request?

dschallis:
Hmm, that might be handy if it's not a feature already :)

It looked like the time parameter was getting converted though, i.e. I was specifying a timestamp down to microsecond precision with:

    'time': '2015-07-27T12:55:48.165231Z'

which was then sent as:

    1438001748165231104

So it looks like it added '104' as nanoseconds from somewhere, instead of sending the value as '1438001748165231', which was the data specified in the 'time' field as a string.

ChristopherRabotin:
@dschallis , just to confirm, the new feature you'd like is to specify a given `time` and the requested precision (e.g. microseconds) and have the influxdb library add the padding, is that correct?
So, for example you would like to pass a time such as `'time': '2015'` and have the library actually log `2015-01-01T00:00:00.0`. Is that correct?

dschallis:
@ChristopherRabotin I think it was more the other way around, i.e. if I specify `'time': '2015-01-01T01:01:01.123456'` and specify precision as seconds, then send data as a timestamp in seconds, instead of as time since epoch in nanoseconds.

ChristopherRabotin:
@aviau , do you reckon this should be a feature to be supported in this client? I think it should be supported at the InfluxDB level directly, especially to avoid having different clients behaving in different ways.

aviau:
[...]

PR #224, opened 2015-08-06
Improve line-protocol

tzonghao:
- Improve DataFrameClient performance under line-protocol, by avoiding datetime parsing.
- Add line-protocol time-precision support (n, u, ms, s)
- Improve line-protocol string concat performance

Partially address #223, when timestamps are integers (n, u, ms, s since EPOCH)
Address #221


aviau:
@tzonghao Thank you for the PR!

PR #227, opened 2015-08-06
Generalize integer timestamp handling

johanfforsberg:
Under some circumstances/platforms/python versions an expected 'int' can actually be of type 'long' (it happened to me on raspberry-pi and I guess it might be because of its 32bit architecture). This change generalizes the handling to accept any integer value.


aviau:
Looks good to me! Thank you for contributing

PR #229, opened 2015-08-15
Added missing database in params.

gallir:
And also: receive an unique point and add it to a list in data["points"] (although it's an unnecessary overhead)

It didn't work at least you provided:

- The database in params['db']
- The dictionary data as the only element of a list in data["points"] which is a nonsense.


gallir:
I will send another pull request because this version failed with write_points.

PR #230, opened 2015-08-16
write() accepts a single point of a list of points

gallir:
This a second part of a previous closed pull request which filed when write() was called from write_points(). Now it checks the structure, if receives a data['points'] its behavior is as before. Otherwise  assumes it's just a point and initializes data['points'] = [data].

So now you can use write() for only one point, it failed before..


aviau:
@gallir Do you want to fix this PR according to @georgijd's comments?

sebito91:
@gallir, just checking if you'd be willing to make the changes referenced above and rebase your commits? If not, we'll look to close this PR.

@aviau update requested.

PR #232, opened 2015-08-27
change line protocol to reflect changes in newest influx

xbs13:
addresses #231

PR #233, opened 2015-08-27
change line protocol to reflect changes in newest influx (now with tests!)

djeebus:
Building off of #232, this fixes the tests and ensures that everything works. @xbs13's code was checking if the value was an int after it had been str'd, so I moved it earlier in the call stack.

On a side note, it appears there was a change in 0.9.3-rc3 (possibly a previous rc) that causes the tags to be returned when calling query. I didn't look into it much, just changed the tests to handle the results coming back.

### DIFF
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -116,6 +116,8 @@ def make_lines(data, precision=None):
             key = _escape_tag(field_key)
             value = _escape_value(point['fields'][field_key])
             if key != '' and value != '':
+                if isinstance(value, int):
+                    value = str(value) + 'i'
                 field_values.append("{key}={value}".format(
                     key=key,
                     value=value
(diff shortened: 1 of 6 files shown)


djeebus:
(the tests are failing b/c it's running against influxdb 0.9.2)

xbs13:
ah! I just saw this after I committed the fix :~

djeebus:
@xbs13 no problem! I got the tests fixed for ya too :)

influxdb/line_protocol.py
```diff
@@ -59,6 +59,8 @@ def _escape_value(value):
                 "\n", "\\n"
             )
         )
+    if isinstance(value, int):
```
xbs13: perhaps `elif isinstance(value, integer_types)` since `six` has that to cover `int`s for py3 and `int`s/`long`s for py2

xbs13:
@djeebus thank you!

influxdb/line_protocol.py
djeebus: Ah, good idea! Will do.

aviau:
@xbs13 @djeebus 

Thank you both! I have merged the changes :)

I'll tag the release today.

PR #236, opened 2015-09-02
Raise InfluxDBServerError for server error instead of InfluxDBClientError

PierreF:
Shouldn't  InfluxDBServerError be raised when InfluxDB fail to serve a request (e.g. HTTP status is 5xx) instead of InfluxDBClientError.

I need this distinction, because I've some code retry writing when server fail (not started/starting) but discard malformed input. E.g. :

```
try:
    client.write_points(...)
except InfluxDBClientError:
    # do not retry this write, it means malformed points
except (InfluxDBServerError, ConnectionError):
   # do retry this write, it means server down/starting
```

This would also be valuable for InfluxDBClusterClient, who should go to next server if one of the server fail.

One case were InfluxDB return a 500 error code (that is not due to bad input) is during startup of single node, before metastore initialization. In this case I got the error ```metastore database error: no leader```. This is obviously only a temporary error and few seconds later server is available.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -202,8 +202,10 @@ localhost:8086/databasename', timeout=5, udp_port=159)
         :type expected_response_code: int
         :returns: the response from the request
         :rtype: :class:`requests.Response`
+        :raises InfluxDBServerError: if the response code is any server error
+            code (5xx)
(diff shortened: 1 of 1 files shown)


aviau:
Thank you for contributing!

PR #243, opened 2015-09-11
Enh/Fix: add a get_free_ports() "enhanced" helper function.

gst:
This fix the problem with get_free_port() used multiple times which can lead to the same port being being reused.


aviau:
Merci Greg :)

PR #244, opened 2015-09-13
Fix issue 237

georgijd:
This should enable the cluster client to properly switch databases and user credentials for all clients in the cluster, not just a random one. It doesn't look pretty and actually I'd like to suggest an alternative implementation of the InfluxDBClusterClient which will actually inherit the InfluxDBClient and instead of maintaining a list of InfluxDBClients, we can just keep the list of (host, port) tuples and alternate those on each request. This way we win a few things:
- Less memory usage, obviously. There's no need to create 3 InfluxDBClient objects when we can get away with a single one;
- We get access to all attributes of the InfluxDBClient  (not just the callable ones);
- Docstrings available for all methods. It also means that auto-completion will be available for developers;
- Easier maintenance and avoiding bugs similar to this one.

Since some users of the library might be using the `clients` attribute of the cluster client current implementation, we can make the new one backwards compatible by having a list with only one element (self) as the `clients` attribute, i.e. ```self.clients = [self]```

@aviau , @gst, @areski , any thoughts on that?

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -763,12 +763,35 @@ class InfluxDBClusterClient(object):
                                                   use_udp=use_udp,
                                                   udp_port=udp_port))
         for method in dir(client_base_class):
-            if method.startswith('_'):
+            if method.startswith('_') or method in ('switch_database',
+                                                    'switch_user'):
                 continue
+
             orig_func = getattr(client_base_class, method)
             if not callable(orig_func):
                 continue
+
             setattr(self, method, self._make_func(orig_func))
+
+    def switch_database(self, database):
+        """Change the database of all clients in the cluster.
+
+        :param database: the name of the database to switch to
+        :type database: str
+        """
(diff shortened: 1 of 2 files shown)


gst:
seems totally legit to me, so +1 :)
Have you a branch already made with what you propose ?

georgijd:
Hi, 
I haven't written anything yet, but I can come up with an initial draft tonight :)

aviau:
Thank you @georgijd

georgijd:
Hey guys, sorry for the delay, it's been a busy week :)
You can check out what I've come up with here: https://github.com/georgijd/influxdb-python/commit/e25576f324c225ff37fd12b1e190b314b8514b45

aviau:
@georgijd Can you transfer this onto this PR?

aviau:
As far as I understand, this is fixed in #253

Issue #237, opened 2015-09-02
switch_database not working with Cluster client


PierreF:
switch_database with clustered client don't work as expected... it only switch database on ONE client :)

For example, the following code:
```
>>> import influxdb
>>> c=influxdb.InfluxDBClusterClient([('localhost', 8086),('localhost', 9086),('localhost', 10086),])
>>> c.create_database('plop')  # success
>>> c.switch_database('plop')  # success (it can't fail :))
>>> c.create_retention_policy('test', '3d', 3, default=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../influxdb/client.py", line 829, in func
    raise e
influxdb.exceptions.InfluxDBClientError: database not found
```
And on InfluxDB console, I see:
```
[http] 2015/09/02 17:31:16 ::1 - root [02/Sep/2015:17:31:16 +0200] GET /query?q=CREATE+RETENTION+POLICY+test+ON+None+DURATION+3d+REPLICATION+3+DEFAULT HTTP/1.1 200 86 - python-requests/2.7.0 CPython/2.7.9 Linux/3.19.0-25-generic a703e054-5187-11e5-8001-000000000000 7.141045ms
```

A simple workaround in my case it to provide database parameter when creating the client. But it would be better if switch_database worked on cluster client, or at least to warn about this issue in doc.

aviau:
This should be fixed in master!

PR #253, opened 2015-10-07
Refactor InfluxDBClusterClient

georgijd:
Use a single client and maintain a list of hosts instead of a list of
clients. Related to #237 and #244


aviau:
Thank you for your work @georgijd. Sorry for the slow response time.

PR #254, opened 2015-10-09
Fixed broken filtering in get_points()

drmclean:
get_points() was not filtering by measurement or tag correctly. It now does.

### DIFF
--- a/influxdb/resultset.py
+++ b/influxdb/resultset.py
@@ -69,6 +69,47 @@ class ResultSet(object):
         return self.get_points(name, tags)
 
     def get_points(self, measurement=None, tags=None):
+        # """
+        # Returns a generator for all the points that match the given filters.
+
+        # :param measurement: The measurement name
+        # :type measurement: str
+
+        # :param tags: Tags to look for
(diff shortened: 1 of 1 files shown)


aviau:
Thank you for contributing!

It looks like you have commented out a whole function, can you take remove it instead?

drmclean:
Thanks for taking the time to reply. 

I've located the source of the failed tests which is that the structure of the response from influxdb has now changed but the response object inside "resultset_test" has not. 

Where previously "tags" was a valid key now they are kept in with columns instead. 
This effects the entire class really including: 

rs.get_points()
rs.items()
rs.keys()

I'm fixing the tests now and will re-submit a pull request once they are complete and passing.

drmclean:
note for posterity. 

Influxdb **used** to include an implicit "group by *" on the end of all requests. This resulted in the explicit inclusion of "tags" in the result which could then be used for filtering. 

Influx then removed this implicit "group by" and so tags are no longer included. This means that ResultSet().get_points() will only filter on measurements or tags if a "group by" is included in the query that is run.

PR #257, opened 2015-10-15
Please add license file to pip package

piotr1212:
text from the mit license:

The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.


aviau:
Thanks!

Issue #258, opened 2015-10-18
FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)


aviau:
@tzonghao Can you take a look at this? It has to do with pandas.

This was reported in Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=802079

tzonghao:
Alex, I won't have access to computer until sometime next week.

________________________________
On: 17 October 2015 17:28, "Alexandre Viau" <notifications@github.com> wrote:

@tzonghao<https://github.com/tzonghao> Can you take a look at this?

-

PR #260, opened 2015-10-26
Fix pandas sort future warning

tzonghao:
Issue #258


aviau:
Thank you! :) I'll be waiting for the build.

tzonghao:
I forgot to base it off from latest master and got some redundant changes in it. I'm fixing it now.

aviau:
Oh that was easy. Thank you for doing it :)!

PR #262, opened 2015-11-07
add option to use Requests http(s) proxy

frennkie:
The influxdb-python client does not have support for a web proxy although
the Requests library provides the possibility to use a web proxy for both
http and https requests.
This PR (optionally) adds the feature to use a http(s) proxy by adding
an additional parameter to InfluxDBClient initialization called´"proxies".
This parameter expects that a dict is provided. If this is not set it will
default to { }.
The syntax for the dict is the same as specified in Requests docs:
http://docs.python-requests.org/en/latest/user/advanced/#proxies
proxies = {
  "http": "http://10.10.1.10:3128",
  "https": "http://10.10.1.10:1080",
}

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -55,6 +55,8 @@ class InfluxDBClient(object):
     :type use_udp: int
     :param udp_port: UDP port to connect to InfluxDB, defaults to 4444
     :type udp_port: int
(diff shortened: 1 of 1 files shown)


aviau:
Thank you for contributing!

PR #265, opened 2015-11-17
Make InfluxDBClusterClient thread-safe

PierreF:
We had an issue with InfluxDBClusterClient an multi-threading : it is not thread safe :(

The thread-safe issues I spotted:

* [random.shuffle](https://github.com/influxdb/influxdb-python/blob/5971d766c8aacf6f5ce5021ea5228cfd97935410/influxdb/client.py#L804) : calling random.shuffle on the same list from multiple thread could corrupt that list
* [removing item from list](https://github.com/influxdb/influxdb-python/blob/5971d766c8aacf6f5ce5021ea5228cfd97935410/influxdb/client.py#L819): two thread may try to remove the same entry from the list
* [update host on ONE client](https://github.com/influxdb/influxdb-python/blob/5971d766c8aacf6f5ce5021ea5228cfd97935410/influxdb/client.py#L810): one thread may update client host before another one called orig_func. Therefor when exception occur or no, the variable "h" may no be the real server used.

For first 2 issues, a simple lock around the affected code solve the issue.

For the last issue, it a bit trickier... I ended with a getter (_get_host() and _get_baseurl()), so InfluxDBClusterClient could override them with code that use a thread local variable.

I've also added properties to avoid changing every access to _host or _baseurl, but since those variables should not be used outside influxdb-python code, we could avoid them and directly use the getter.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -6,6 +6,7 @@ Python client for InfluxDB
 from functools import wraps
 import json
 import socket
+import threading
 import random
 import requests
 import requests.exceptions
@@ -73,7 +74,7 @@ class InfluxDBClient(object):
                  proxies=None,
                  ):
         """Construct a new InfluxDBClient object."""
-        self._host = host
+        self.__host = host
         self._port = port
         self._username = username
         self._password = password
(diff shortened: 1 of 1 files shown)


aviau:
thank you for your work!

PR #267, opened 2015-11-19
Added healing feature on InfluxClusterClient

PierreF:
With a long running process, the Cluster client had an issue: once a server is moved to bad_hosts, it's only retried when all other hosts also failed. So when it recovers, only a subset of healthy server are used.

The worse case, it when all server was down,  as soon as one server if back online, it become the only server used.

This pull request fix this issue by moving server from bad_hosts to healthy hosts list after a delay.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -6,6 +6,7 @@ Python client for InfluxDB
 from functools import wraps
 import json
 import socket
+import time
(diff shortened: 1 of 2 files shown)


aviau:
Thanks! This is a feature I was hoping for.

PR #271, opened 2015-11-27
Fix create/drop numeric-named databases

slomek:
This should fix https://github.com/influxdb/influxdb-python/issues/270


aviau:
Thanks for the tests!

PR #276, opened 2015-12-05
Create db only if doesn't exist yet

slomek:
This is a follow-up of https://github.com/influxdb/influxdb-python/pull/274

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -454,13 +454,16 @@ localhost:8086/databasename', timeout=5, udp_port=159)
         """
         return list(self.query("SHOW DATABASES").get_points())
 
-    def create_database(self, dbname):
+    def create_database(self, dbname, if_not_exists=False):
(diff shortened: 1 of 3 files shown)


aviau:
@slomek Have you taken a look at the tests failures?

slomek:
Yep, I've had some issues with running them locally. I've changed the default behaviour to create dbs if they do not exist, but also added a flag if you want to fail anyway (and not to delete one of your tests). This breaks backward compatibility a little, but in my opinion it would be better this way.

slomek:
@aviau Can you merge this or is there anything else I should fix first?

aviau:
@slomek I am sorry for the long response time. Can you please switch the default around? Some people may rely on this throwing exceptions and I think its a saner default.

I will merge right after!

slomek:
No problem, @aviau. Take a look on it now.

aviau:
Thanks!!

PR #277, opened 2015-12-08
Use urlparse for parse netloc

william-p:
IMO regex is a bad idea for parse something like this.
In this PR i use `urlparse`.

What do you think ?


JulienChampseix:
+1

aviau:
Thanks!

PR #279, opened 2015-12-13
Improve ResultSet.items() performance

a-sk:
Use straightforward method of getting points for a serie:
no need to try to filter by measurement/tag: it greatly decreases performance.

In my case performance of .items() for ResultSet over 850 series with ~10000 points decreases from ~1.5 seconds to ~30ms


aviau:
Sorry for the delay, and thank you for contributing @a-sk !

a-sk:
Thanks!

PR #283, opened 2016-01-08
Python 2.6 does not support keyword arguments in decode

adrianlzt:
http://docs.python.org/library/stdtypes.html#str.decode
Changed in version 2.7: Support for keyword arguments added.


aviau:
Thanks!

PR #287, opened 2016-01-26
Fix issue with cluster on multiple ports

PierreF:
On PR #265, the changes added to make InfluxDBClusterClient thread-safe only covered host and baseurl, port was not covered.

This cause an issue when your cluster is using different ports : only port of the first server is used.


aviau:
I have restarted failing jobs.

aviau:
Failing tests were fixed on master. This should work.

PR #291, opened 2016-01-30
create_user() - added quotes to the username

mkren:
Created because of this issue:
This code does not work with "error parsing query":
client.create_user('aaa-bbb', 'pass', admin=False)


aviau:
thank you for the PR!

PR #295, opened 2016-02-17
change package name to avoid conflict

sharang:
use `influxdb` will cause conflict in CentOS/RedHat yum

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ with open('README.rst', 'r') as f:
 
 
 setup(
-    name='influxdb',
+    name='influxdb-python',
     version=version,
     description="InfluxDB client",
     long_description=readme,


sharang:
@aviau thoughts?

aviau:
What? What kind of conflict?

sharang:
@aviau please see the last line 

```shell
[root@xy-dev influxdb-python]# rpm -ivh ~/sandbox/3parties/setuppackets/deps/influxdb-0.10.0-1.x86_64.rpm 
Preparing...                ########################################### [100%]
   1:influxdb               ########################################### [100%]
[root@xy-dev influxdb-python]# rpm -q influxdb
influxdb-0.10.0-1.x86_64
[root@xy-dev influxdb-python]# python setup.py bdist_rpm 
[root@xy-dev influxdb-python]# rpm -i dist/influxdb-2.12.0-1.noarch.rpm 
[root@xy-dev influxdb-python]# rpm -q influxdb
influxdb-0.10.0-1.x86_64
influxdb-2.12.0-1.noarch

```

sharang:
@aviau if the name in setup.py is `influxdb`, the distro package name is also `influxdb-xxxx.rpm`

aviau:
The proper naming for python libraries it python-influxdb. You generated a package with a dirty script, please just name the RPM properly.

sharang:
@aviau  bug I just use the `setup.py` to generate rpm.

`[root@xy-dev influxdb-python]# python setup.py bdist_rpm `

sharang:
@aviau  as you said, `The proper naming for python libraries is python-influxdb`, but why the `name` in  setup.py is `influxdb` ...

aviau:
The package name is different than the module name.

For example, requests's package name is python-requests but the [module name is requests](https://github.com/kennethreitz/requests/blob/master/setup.py#L63)

sharang:
@aviau so, how to build rpm package with prefix `influxdb-python`?

aviau:
Take a look at [this guide](https://fedoraproject.org/wiki/How_to_create_an_RPM_package). Or some other guide. This repository has nothing to do with distro packaging. I happen to be the Debian maintainer for influxdb-python, but I maintain in another repository in the Debian infrastructure.

sharang:
@aviau but why not change it to make more clear?

currently the document said:

```
Install, upgrade and uninstall InfluxDB-Python with these commands:

$ pip install influxdb
$ pip install --upgrade influxdb
$ pip uninstall influxdb
On Debian/Ubuntu, you can install it with this command:

$ sudo apt-get install python-influxdb
```

why the name in pip command is `influxdb`, while in apt-get command is `python-influxdb`.
And if people use `python setup.py bdist_rpm` they will get a package named `influxdb`.

sharang:
@aviau `influxdb` usually means the database, while this project is the python client, so I think it is better to change the name.

It is different from `requests`, since there is no other `requests`, event if you google `requests`, the python lib will be the first one.

sharang:
@aviau  package name can be different from module name, but why not make it the same :)

aviau:
> why the name in pip command is influxdb, while in apt-get command is python-influxdb.

Because that is what every other package does. ALL of them. From http://pypi-ranking.info/alltime
 - simplejson
 - setuptools
 - six
 - requests
 - virtualenv
 - [...]

__NONE__ of them have a ```python-``` prefix in the module name. And __ALL__ of them have the ```python-``` prefix in the distro package name.


> @aviau package name can be different from module name, but why not make it the same :)

Because I am following standards.

sharang:
@aviau 

```      
[root@xy-dev test]# ls py
pycrypto-2.6.1/        pycrypto-2.6.1.tar.gz  pylint-mode/           pymongo-3.0.3/         pymongo-3.0.3.tar.gz   pystatsd/              pyzmq/                 
[root@xy-dev test]# grep -wE "('name':)|(name=)" pycrypto-2.6.1/setup.py 
kw = {'name':"python-crypto",
[root@xy-dev test]# grep -wE "('name':)|(name=)" pymongo-3.0.3/setup.py 
    name="pymongo",
[root@xy-dev test]# grep -wE "('name':)|(name=)" pystatsd/setup.py 
    name='pystatsd',
[root@xy-dev test]# grep -wE "('name':)|( name = )" pyzmq/setup.py 
    name = "pyzmq",
```

things are different, there are no other `simplejson, setuptools, six, requests`, but there are `crypto, mongo, statsd, zmq`...

sharang:
@aviau and `MySQL-python`, `pycparser`, `pytz`, `PyYAML`, `pyasn1`  From http://pypi-ranking.info/alltime

aviau:
> @aviau and `MySQL-python`, `pycparser`, `pytz`, `PyYAML`, `pyasn1`  From http://pypi-ranking.info/alltime

The distro package name for those packages is still different. For example the debian package for `pyasn1` is ```python-pyasn1```. So they don't address the issue you are pointing to.

sharang:
@aviau if you see database, there are `py-postgresql`, `PyHBase`.

What I mean is that, `influxdb` is already mean something, so when we use `python setup.py bdist_rpm` to build rpm for this project, we should better generate rpm with another name...

sharang:
@aviau If we can not use the `standard` generate command (`python setup.py bdist_rpm`), the document should tell people what they can use.

aviau:
> @aviau If we can not use the standard generate command (python setup.py bdist_rpm), the document should tell people what they can use.

Feel free to open a pull request. I am locking this issue.

PR #296, opened 2016-02-17
Fix unicode data handling in tag's content

anonymouzz:
I'm discovered that client crashes in make_lines function if unicode data in tag content (under py27).
This patch fix it for me.

### DIFF
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from numbers import Integral
 
 from dateutil.parser import parse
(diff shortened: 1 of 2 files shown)


aviau:
Great! Thank you for your work.

caioariede:
Do you mind releasing a new version w/ this on PyPI?

It's pretty critical for anyone handling unicode data in Python 2.7.

pengyao:
+1 for a new viesion

AlexGrs:
+1 also. As a workaround, i forked master for my own need but would prefer to keep using main package.

PR #301, opened 2016-02-28
Make InfluxDBClient to force an integer as the HTTP(s) port.

canardleteer:
- Includes unit tests
- Does not make the change to influxdb08
- Also a small change that was needed to tox.ini that was needed for tox to pass, since setup.py reads README.rst


aviau:
I launched tests again, but otherwise it all looks good! I'll wait for travis to pass.

canardleteer:
@aviau - resolved the tox thing, i did the development on my fairly outdated laptop environment. All passing now.

PR #304, opened 2016-03-08
Allow setting the time of a point manually.

### LINKED ISSUE #259: How to set time using SeriesHelper?
I tried to use SeriesHelper to set the time, but it keeps getting overridden with default current time when saving into DB. I have added it to the fields (don't know if this is correct, as the documentation does not specify anything about setting time):

```
class MeteorologicalDataHelper(SeriesHelper):
    # Meta class stores time series helper configuration.
    class Meta:
        # The client should be an instance of InfluxDBClient.
        client = myclient
        # The series name must be a string. Add dependent fields/tags in curly brackets.
        series_name = 'meteorological_data'
        # Defines all the fields in this time series.
        fields = ['time', 'air_temperature']
        # Defines all the tags for the series.
        tags = []
        # Defines the number of data points to store prior to writing on the wire.
        bulk_size = 100
        # autocommit must be set to True when using bulk_size
        autocommit = True
```

But it seems that this value is not used, for example I have executed the following:
`MeteorologicalDataHelper(air_temperature=12, time=1434067467000000000)`

And then checked in console of influx and the time was set to current time instead of the provided one:
```
> select * from meteorological_data
name: meteorological_data
-------------------------
time				air_temperature
2015-10-25T13:05:06.736935489Z 	12
```

### LINKED ISSUE #264: SeriesHelper inserting only last point from the set
I am using influxdb 0.9.4.2
and this example to submit multiple points to database
http://influxdb-python.readthedocs.org/en/latest/examples.html#tutorials-serieshelper

I am trying to post 35 points per query.
I don't know if it's just me, but whenever I run my script it adds to db only last point.
If I set bulk_size=1 and in python script insert time.sleep(1) after every MySeriesHelper() call
then I am able to post most of the points(32-35).

grundleborg:
Point can be specified as either a number of nanoseconds, a python
datetime object (with or without timezone) or a string in ISO datetime
format.

If a time is not specified, the Helper sets the time at the time of
assembling the point fields so that multiple unique points with the same
tags can be committed simultaneously without them failing to add due to
all being assigned the same automatic time by the InfluxDB server.

This fix is based upon the discussion in #130 but also includes the
outstanding items for it to be merged. I'm happy to receive suggestions
for further ways to add test coverage to this change.

This also fixes #264 and fixes #259.

### DIFF
--- a/influxdb/helper.py
+++ b/influxdb/helper.py
@@ -3,6 +3,7 @@
 Helper class for InfluxDB
 """
 from collections import namedtuple, defaultdict
+from datetime import datetime
 from warnings import warn
 
 import six
@@ -15,6 +16,16 @@ class SeriesHelper(object):
     All data points are immutable, insuring they do not get overwritten.
     Each subclass can write to its own database.
     The time series names can also be based on one or more defined fields.
+
+    A field "time" can be used to write data points at a specific time,
+    rather than the default current time. The time field can take any of
+    the following forms:
+     * An integer unix timestamp in nanoseconds, assumed to be in UTC.
+     * A string in the ISO time format, including a timezone.
+     * A naive python datetime, which will be treated as UTC.
+     * A localized python datetime, which will use the chosen timezone.
+    If no time field is provided, the current UTC system time in microseconds
+    at the time of assembling the point data will be used.
 
     Annotated example::
 
@@ -142,8 +153,23 @@ class SeriesHelper(object):
                     "tags": {},
                 }
 
+                ts = getattr(point, 'time', None)
+                if not ts:
+                    # No time provided. Use current UTC time.
+                    ts = datetime.utcnow().isoformat() + "+00:00"
+                elif isinstance(ts, datetime):
+                    if ts.tzinfo is None or ts.tzinfo.utcoffset(ts) is None:
+                        # Assuming naive datetime provided. Format with UTC tz.
+                        ts = ts.isoformat() + "+00:00"
+                    else:
+                        # Assuming localized datetime provided.
+                        ts = ts.isoformat()
+                # Neither of the above match. Assuming correct string or int.
+                json_point['time'] = ts
+
                 for field in cls._fields:
-                    json_point['fields'][field] = getattr(point, field)
+                    if field != 'time':
+                        json_point['fields'][field] = getattr(point, field)
 
                 for tag in cls._tags:
                     json_point['tags'][tag] = getattr(point, tag)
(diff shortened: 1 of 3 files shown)


matt-snider:
Hi @grundleborg!

I've just opened a PR for the exact same feature. I didn't see that yours already existed, and didn't think to check because I had already told @aviau that I would work on this feature a few weeks ago. Had I seen this PR I would have discussed it with you beforehand. 

I've taken a slightly different approach as the one in your PR & #130 that avoids a few issues:
   * Assigning a default time in `_json_body()` will result in a race condition where the points will have a very similar and possibly even identical timestamp, meaning in some cases #264 can still occur
      * Instead, I've done this at point-creation which also has the benefit that the default timestamps are more accurate
   * By rolling your own timestamp conversion, you duplicate some logic in [line_protocol.py](https://github.com/influxdata/influxdb-python/blob/master/influxdb/line_protocol.py#L13) for handling timestamps 

Take a look and let me know what you think :smiley:. Maybe @aviau can also weigh in.

grundleborg:
@matt-snider no problem, and quite happy for yours to be merged instead if it handles things more robustly. I needed the feature right away so couldn't wait for someone else to implement it, and I thought I might as well send a PR in since I didn't know how long it would be before anyone else actually dealt with it (and the original PR where the discussion was happening didn't seem to be going anywhere fast).

matt-snider:
Yep, I understand, especially considering the original PR was opened nearly a year ago...

The important part is that we get this feature merged! :smile:

aviau:
Thank your for your contribution!

matt-snider:
Hi @aviau, did you even take a look at my PR? As I mentioned, I don't think this accurately solves #264.

aviau:
It does. You are setting a time when the point is created.

matt-snider:
Correct me if I'm wrong but won't all points receive approximately the same timestamp, as they are generated only when `_json_body()` is called. So depending on how quickly the loop runs, the precision used, etc, they may end up being sent over the wire with an identical timestamp.

aviau:
The timestamps were originally equivalent because InfluxDB was setting one when receiving the batch. I don't know how quickly _json_body() runs but I wouldn't expect timestamps to be the same. We can also move the timestamp generation to when the user creates the point from the serieshelper.

matt-snider:
This is exactly the approach I took because it results in more accurate timestamps and rules out any chance of accidental loss of data (i.e. due to a race condition).

Look I'm not advocating my approach over @grundleborg 's just because I want to have my code merge or anything like that. But if we're implementing automatic timestamps, shouldn't they be accurate to some degree? What about the case where I'm using a batch size (e.g. 500) and those points now all end up with the timestamp generated immediately before being sent?

PR #306, opened 2016-03-13
Add time parameter to SeriesHelper

matt-snider:
This PR replaces #130 and takes into account the feedback from that discussion.

This will allow the time to be specified when creating a point with `SeriesHelper`. This can be done regardless of whether *time* was declared as a field in the `SeriesHelper.Meta` class. 

If time is not specified the implementation defaults to using `datetime.utcnow()`. This occurs at point-creation, not within the call to `_json_body_()`. This is important because to solve #264/#259 successive points within the same series should still be retained. If the time is recorded in `_json_body_()`, the points end up all having approximately the same time and you end up with a race condition and inaccurate timestamps.

As I've added to the docstring for `SeriesHelper`, the time can be any of the time types supported by the client (see [line_protocol.py#L13](https://github.com/influxdata/influxdb-python/blob/master/influxdb/line_protocol.py#L13)).  A nice thing about using a `datetime` object as the default, is that this leverages the existing precision handling in `_convert_timestamp()`.

One area on which I'd like feedback, is whether it's undesirable that if 'time' is specified as a field, but not specified at point-creation, that it won't raise a `NameError` as per the usual logic for missing fields/tags in [helper.py#L104](https://github.com/influxdata/influxdb-python/blob/master/influxdb/helper.py#L104).

### DIFF
--- a/influxdb/helper.py
+++ b/influxdb/helper.py
@@ -6,6 +6,7 @@ from collections import namedtuple, defaultdict
 from warnings import warn
 
 import six
+import time
 
 
 class SeriesHelper(object):
@@ -87,8 +88,11 @@ class SeriesHelper(object):
                         ' autocommit is false.'.format(cls.__name__))
 
             cls._datapoints = defaultdict(list)
-            cls._type = namedtuple(cls.__name__, cls._fields + cls._tags)
 
+            if 'time' in cls._fields:
+                cls._fields.remove('time')
+            cls._type = namedtuple(cls.__name__,
+                                   cls._fields + cls._tags + ['time'])
         return super(SeriesHelper, cls).__new__(cls)
 
     def __init__(self, **kw):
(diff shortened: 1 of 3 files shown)


matt-snider:
It looks like the py26 build failed due to the pandas installation failing. 

I can't see anything I changed that would affect this, but pandas 0.18.0 was released two days ago, so maybe it contained a breaking change.

nelg:
will this support sending the time including microseconds?   ie, with a precision option?

matt-snider:
Hey @nelg - take a look at #304 which was merged instead of this one.

matt-snider:
@aviau As discussed I updated my PR. 

The tests all pass except for py26 but it looks like this has already been happening in the last few travis builds. Could this have something to do with [pandas 0.18.0](https://github.com/pydata/pandas/releases) dropping python2.6 support?

Let me know what you think.

aviau:
@matt-snider Thank you for your patience and thank you for the PR.

This project needs some love, should you have time to work on more issues, I would welcome your help!

matt-snider:
@aviau No problem - I'm glad to help! I'll hopefully have some more time in the near future to contribute a bit more :smile:

Issue #320, opened 2016-05-13
[Bug] Retention policy functionality doesn't escape/backtick policy name or database preventing legitimate characters


afelle1:
The following retention policy commands in [influxdb-python/influxdb/client.py](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py) do not escape/backtick database and/or retention policy name thus breaking when allowed characters are used elsewhere (e.g. database names containing hyphens like "db_0474102f-8397-481d-bcba-06f3a682854d").

Note: This might extend to more than simply retention policy commands but will allow maintainers to determine scope as it should be universally done.

1. [create_retention_policy](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L483)
   ```

       query_string = \
            "CREATE RETENTION POLICY %s ON %s " \
            "DURATION %s REPLICATION %s" % \

   ```
2. [alter_retention_policy](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L514)
   ```

        query_string = (
            "ALTER RETENTION POLICY {0} ON {1}"

   ```
3. [drop_retention_policy](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L551)
   ```

        query_string = (
            "DROP RETENTION POLICY {0} ON {1}"

   ```
4. [get_list_retention_policies](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L565)
   ```

        rsp = self.query(
            "SHOW RETENTION POLICIES ON %s" % (database or self._database)

   ```

PR #321, opened 2016-05-13
Updated client retention policy functionality to escape/backtick database and policy name

afelle1:
Original bug reported https://github.com/influxdata/influxdb-python/issues/320


aviau:
Thanks!

afelle1:
∑=3 looks like the tests are being ran against influxdb 0.13, which made changes to a number of different commands, so the tests are having a bad time.  Any thoughts about what to do there?

aviau:
We will have to update all of the commands to the latest version of InfluxDB.A Are you willing to help?

PR #322, opened 2016-05-13
add grant listing and granting admin

aisch:
needed for an ansible `influxdb_user` (make admin) and `influxdb_privs` (check if priv granted) modules

PR #330, opened 2016-06-08
Add get_list_privileges() to InfluxDBClient

babilen:
This allows library users to get a list of all privileges an InfluxDB
user has been granted.

One problematic aspect of this method is that, due to https://github.com/influxdata/influxdb/issues/4511 the interface of privilege related methods is inconsistent.

We could perform the normalisation (removal of elements with`NO PRIVILEGES` and `ALL PRIVILEGES` → `ALL`) in the Python library, but I would prefer to keep the library simple and to see this fixed this in InfluxDB itself.

On the other hand including it would probably lead to client code that performs the normalisation itself and that would have to be adapted once these inconsistencies have been resolved.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -728,6 +728,28 @@ localhost:8086/databasename', timeout=5, udp_port=159)
                                                    username)
         self.query(text)
 
+    def get_list_privileges(self, username):
+        """Get the list of all privileges granted to given user.
+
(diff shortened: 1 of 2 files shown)


babilen:
I just noticed that this PR is pretty similar to #322, but I prefer to use `get_list_privileges` over `get_list_grants` as this matches the existing interface and documentation more closely.

Issue #323, opened 2016-05-13
travis tests have been failing for a while


aisch:
seems since ~ https://travis-ci.org/influxdata/influxdb-python/builds/116902069

aviau:
We should fix this ASAP before merging new things.

babilen:
@aviau A lot of these tests seem to fail as points are now unicode strings rather than str. A possible fix would be to adapt the tests to it or to figure out what introduced that change.

What would you see as the preferred approach?

aviau:
Adapting the tests is fine as long as they show that the library works ;)

babilen:
Well, I am not a huge fan of simply adapting the test suite so that the tests pass. Something must have triggered the type change and in some tests types seem to be inconsistent.

See, for example, the following:

```
======================================================================
FAIL: test_get_list_series_and_delete (influxdb.tests.server_tests.client_test_with_server.CommonTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/babilen/src/github.com/influxdb/influxdb-python/influxdb/tests/server_tests/client_test_with_server.py", line 480, in test_get_list_series_and_delete
    rsp
AssertionError: Lists differ: [{u'name': u'cpu_load_short', ... != [{'name': 'results', 'tags': [...

First differing element 0:
{'name': 'cpu_load_short', 'tags': [{'_key': 'cpu_load_short,host=server01,region=us-west', u'region': u'us-west', u'host': u'server01'}]}
{'name': 'results', 'tags': [{u'key': u'cpu_load_short,host=server01,region=us-west'}]}
```

where we see a mixture of `unicode` and `str`.

aviau:
We should return unicode to the user everywhere if that is possible.

babilen:
Do you have an idea why points and tags behave that way, while `'name': 'results', 'tags'` in there do not?

aviau:
We probably create those keys in result_set.py.

babilen:
Most problems stem from the upgrade to InfluxDB 0.12.1 and the resulting changes in 103c7dc85607b3bb157af4cca4972c006bae7780

babilen:
Updating to `0.13.0` lets `test_issue_143` pass again due to https://github.com/influxdata/influxdb/pull/6283

aviau:
Thanks to @babilen, the test suite was fixed!

PR #331, opened 2016-06-09
Add grant_admin_privileges() to InfluxDBClient

babilen:
This allows users to easily alter the 'admin' attribute of a user
account and complements the already existing 'revove_admin_privileges'
method.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -684,6 +684,18 @@ localhost:8086/databasename', timeout=5, udp_port=159)
                                                    for k, v in tags.items()])
         self.query(query_str, database=database)
 
+    def grant_admin_privileges(self, username):
+        """Grant cluster administration privileges to a user.
+
+        :param username: the username to grant privileges to
+        :type username: str
+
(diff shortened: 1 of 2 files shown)


aviau:
@babilen  I'd love to see the tests passing before merging this. I tank you for your contribution, would you like to open a PR to fix the tests?

babilen:
The tests seem to be failing due to changes introduced in https://github.com/influxdata/influxdb-python/commit/f97effe031dd61508faa0f8f699099a514225b9e asmentioned in https://github.com/influxdata/influxdb-python/issues/323. I don't think that is specific to my changes here.

aviau:
You are right, but the fact that there are failing tests makes reviewing pull requests longer for me because I have to read the logs and see what is failing and what isn't.

babilen:
Sure, I totally understand. It's just that https://github.com/influxdata/influxdb-python/commit/f97effe031dd61508faa0f8f699099a514225b9e introduced significant changes and I am, as of now, not entirely sure how to fix it.

I will, however, take a look, but I am not yet too familiar with the codebase.

Issue #329, opened 2016-06-08
Queries NOT SELECT and SHOW should be posts


wx00:
Since v0.13, queries sent to the database `/query` endpoints that are not `SELECT` or `SHOW` should be using `POST` rather than `GET`, and v1.0.0 is now in beta, this should be fixed recently.

Ref: 

[Deprecated `/query` endpoint request behavior](https://docs.influxdata.com/influxdb/v0.13/administration/012_vs_013/#deprecated-query-endpoint-request-behavior)

[query `GET` in source code](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L329)

aviau:
I don't have much time to work on this right now, but I am able to review and merge if someone wants to do the work.

sebito91:
Closing stale issue, please reopen if this is still required.

PR #334, opened 2016-06-13
Updated query method for v0.13 and later server.(#329)

wx00:
`query` now uses `GET` for  `SELECT` or `SHOW` queries, and `POST` for others.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -328,9 +328,13 @@ localhost:8086/databasename', timeout=5, udp_port=159)
         if epoch is not None:
             params['epoch'] = epoch
 
+        method = 'POST'
+        if query.split() and query.split()[0].upper() in ('SELECT', 'SHOW'):
+            method = 'GET'
+
         response = self.request(
             url="query",
-            method='GET',
+            method=method,
             params=params,
             data=None,
             expected_response_code=expected_response_code
--- a/influxdb/tests/client_test.py
+++ b/influxdb/tests/client_test.py
@@ -408,7 +408,7 @@ class TestInfluxDBClient(unittest.TestCase):
     def test_create_database(self):
         with requests_mock.Mocker() as m:
             m.register_uri(
-                requests_mock.GET,
+                requests_mock.POST,
                 "http://localhost:8086/query",
                 text='{"results":[{}]}'
             )
@@ -421,7 +421,7 @@ class TestInfluxDBClient(unittest.TestCase):
     def test_create_numeric_named_database(self):
         with requests_mock.Mocker() as m:
             m.register_uri(
-                requests_mock.GET,
+                requests_mock.POST,
                 "http://localhost:8086/query",
                 text='{"results":[{}]}'
             )
@@ -439,7 +439,7 @@ class TestInfluxDBClient(unittest.TestCase):
     def test_drop_database(self):
         with requests_mock.Mocker() as m:
             m.register_uri(
-                requests_mock.GET,
+                requests_mock.POST,
                 "http://localhost:8086/query",
                 text='{"results":[{}]}'
             )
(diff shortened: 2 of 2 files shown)


babilen:
This would fix https://github.com/influxdata/influxdb-python/issues/329

wx00:
Thanks.
I found it brokes test cases, and I'm now trying to update those test queries that are not `SELECT` or `SHOW` with `GET` method.
May update tonight or tomorrow (using China Standard Time).

babilen:
Which test cases are these?

wx00:
I think it's ALL queries tests except those using `SELECT` and `SHOW` queries. So, a lot of the tests should be modified...

babilen:
I didn't see a difference in tests after cherry picking your commit. It's still a needed change though :)

wx00:
I found the problem after sending this PR, didn't make changes in tests yet, will do it later in a few days..

wx00:
@babilen @aviau, can you help me with the test cases? Looks like the influxdb v0.12.1 doesn't support `POST` method in some queries.

babilen:
Please rebase your branch against master and force push to GH. That will include a number of fixes that are necessary to get this running. In particular it will ensure that InfluxDB 0.13 is installed.

wx00:
I modified client_test.py and it passed all tests in py2.7 on my test machine now. Do I need to send another pull request after update?

wx00:
conflicts [here?](https://github.com/influxdata/influxdb-python/compare/master...wx00:master#diff-58c51d2852377954867cd17a8b327af0R718)

babilen:
The problem is that you developed in your "master" branch and didn't rebase when you pulled in changes. I would generally recommend to create a specific feature branch (e.g. `git checkout -b query-method`), pull upstream changes into master (`git pull` in the master branch) and rebase your branch against master every now and then (`git rebase master` in your `query-method` branch).

Your master branch is in some strange state right now and would probably have to be reset. I checked your commit and the tests pass and it applies cleanly though, so the question is how to clean up this PR.

Maybe a `git pull --rebase` is all you need in your master branch, but maybe you have to create the feature branch, hard reset your master branch to some earlier commit and rebase your feature branch against master with `git rebase master`.

babilen:
Would `query.split()[0].upper()` work with an empty query? It looks as if that would throw an `IndexError` that doesn't make it easy for the user to understand what went wrong.

It would be nice if you could default to one method and then override it *if* a query has been passed and if it starts with either `SELECT` or `SHOW`. We could also think about using `.startswith(...)` here.

wx00:
Good point, I didn't think of empty string, that would throw an `IndexError` as you said.
The reason I use `split()` here is, if the query is start with whitespace character, it can still get the first word.
It looks like `.startswith(...)` can only match one word at a call? 

How do you think about this code?
``` python
method = 'POST'
if query.split() and query.split()[0].upper() in ('SELECT', 'SHOW'):
    method = 'GET'
```

wx00:
I'm not so familiar with all those git branch operations, I'll try to fix it, thanks for the help.

babilen:
Wonderful, thank you for looking into this. `.startswith(...)` would only be able to look at one string at a time, yeah.

I personally wouldn't mind:

```
method = 'POST'
normalised_query = query.lower()
if normalised_query:
  if query.startswith('select'):
    method = 'GET'
  elif query.startswith('show'):
    method = 'GET'
```

Or parsing the query explicitly like you do. I simply don't like chaining multiple method calls and index lookups in the same statement (things like `split()[0].upper()`) and tend to split it into multiple steps.

@aviau What do you think would be a good approach here?

wx00:
Hmm..
 I hard reset `master` and pulled from upstream, and it's synced with upstream then(check by `git log`), then I apply a patch from my previous "future" commit. So the codes should be fine and now with a clean commit history.
But I just don't sure how it will behave when I push onto github... Is it OK to push?

wx00:
Your method here still has the problem when the query starts with whitespace characters and then `select` or `show`..
No problem if it's no characters in string, it will be dealt in the later `request(...)`.

babilen:
You can't break anything by force pushing into your own repository and it sounds as if you have a clean branch now.

I agree that the approach I sketched above does have issues too. I simply wanted to sketch the idea, rather than provide a full implementation. Basically just wanted to ensure that the data handling is robust.

rezen:
Can we get this merged in?

wx00:
Outdated

Issue #344, opened 2016-06-20
Timestamps slightly different in Influxdb


hirochri:
Occasionally my data arrives in the database with timestamps that are ever so slightly ahead or behind the time they actually should be, and I have been having trouble tracking down where this behavior is coming from!

I am using a pandas dataframe with a datetime index, which is being written to the database using [write_points](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L352) (No specified time precision). The timestamps within the dataframe are numpy datetime64[ns]'s and Influx's default time precision is also nanoseconds iirc. What often happens though is that the timestamps in the database are slightly different and looks like it possibly has to do with floats? (My problem is similar to [this](https://groups.google.com/forum/#!topic/influxdb/uyp1kmMXwxI), as queries based on time now miss 1 or 2 data points)

For instance, this is how the following timestamps change:

datetime64[ns]                              -->         unix timestamp in Influxdb
2016-06-20 13:35:42.270000       -->        1466429742270000128
2016-06-20 14:02:49.770000       -->        1466431369769999872

If anyone could clarify why this behavior is happening or point me in the right direction of how to fix it, that would be greatly appreciated, thanks!

Edit: After glancing at the other possible issues, it may very well be related to [Issue 340](https://github.com/influxdata/influxdb-python/issues/340)

inselbuch:
this is still a problem... it is simply not possible to reliably overwrite data by specifying the exact same timestamp

nargiza-sarkulova:
I also experienced this problem. I'm very much interested in answer.

Issue #340, opened 2016-06-14
dataframe client precision "ns" causing round error


jevans121:
It appears, that the line,
_dataframe_client.py:145 
             'time': int(ts.value / precision_factor)

Is causing a time rounding error when precision_factor is 'ns'. It is likely to be insufficient precision in the double64 type to hold this many seconds (inc 9 decimal places). The rounding error is in the order of (<100ns). This is consistent with the 15 sig.figures of the double64.

The issue is not present when using the line without the division e.g.:
             'time': ts.value

but only for precision_factor, None or 'ns'. Other precisions (unsurprisingly) create time out of bounds on the POST to influxdb.

This line was added at commit:
https://github.com/influxdata/influxdb-python/commit/02ddf0e783d57762341fbf0662ac3b6c9696dcbc

PR #346, opened 2016-06-21
Adjustments to deal with time precision rounding errors in _convert_dataframe_to_json()

hirochri:
I believe this fixes issues [344](https://github.com/influxdata/influxdb-python/issues/344) and [340](https://github.com/influxdata/influxdb-python/issues/340) with timestamps rounding weirdly when nanosecond (or None) time precision is specified


sebito91:
Thank you so much for contributing this PR @hirochri but need to close this in lieu of more recent changes to the codebase.

Issue #342, opened 2016-06-16
Exception on get_list_servers()


tejasbehra:
Influxdb 0.13v 

Traceback (most recent call last):
  File "store_log.py", line 8, in <module>
    print client.get_list_servers()
  File "/usr/local/lib/python2.7/dist-packages/influxdb/client.py", line 615, in get_list_servers
    return list(self.query("SHOW SERVERS").get_points())
  File "/usr/local/lib/python2.7/dist-packages/influxdb/client.py", line 325, in query
    expected_response_code=expected_response_code
  File "/usr/local/lib/python2.7/dist-packages/influxdb/client.py", line 248, in request
    raise InfluxDBClientError(response.content, response.status_code)
influxdb.exceptions.InfluxDBClientError: 400: {"error":"error parsing query: found SERVERS, expected CONTINUOUS, DATABASES, DIAGNOSTICS, FIELD, GRANTS, MEASUREMENTS, QUERIES, RETENTION, SERIES, SHARD, SHARDS, STATS, SUBSCRIPTIONS, TAG, USERS at line 1, char 6"}

PR #356, opened 2016-07-30
Remove get_list_servers()

gaker:
Fix influxdata/influxdb-python#342

``SHOW SERVERS`` was removed in influxdata/influxdb#6470


aviau:
Right, this should leave the codebase.

Thank you @gaker !

PR #360, opened 2016-08-11
make session a parameter

dschien:
I would like to reuse the session while using the client in my django app. For this purpose I have made the session a parameter.


aviau:
This will create a session for all future clients.

You must do something like

```
function(session=None):
    if session is None:
        session=session.Session()
```

PR #361, opened 2016-08-15
Properly quote all identifiers and literals in InfluxDBClient

saaros:
Previously InfluxDBClient tried to quote some of the identifiers (database
and usernames, etc), but a number of API calls didn't have any quoting, and
the ones that had quoting didn't account for special characters inside the
identifiers.

Add new `quote_ident()` and `quote_literal()` functions to line_protocol.py
and use them consistently in the client.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -13,7 +13,7 @@ import requests
 import requests.exceptions
 from sys import version_info
 
-from influxdb.line_protocol import make_lines
+from influxdb.line_protocol import make_lines, quote_ident, quote_literal
 from influxdb.resultset import ResultSet
 from .exceptions import InfluxDBClientError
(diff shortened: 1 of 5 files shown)


wladekb:
This PR may help with such obvious issues like this one:
```
ic.write_points([{
  'measurement':'t1',
  'fields':{
    'a':'a\\' // HERE: slash at the end of the field value
  }
}])
```
which apparently ends up with broken line and influxdb returns 400 Bad Request (unbalanced quotes)

Any chance that this could be merged soon?

aviau:
Thanks!

virusdefender:
This PR fixed my issue https://github.com/influxdata/influxdb-python/issues/365

thanks @saaros

lwis:
When will this be released?

PR #367, opened 2016-09-09
Fix DataFrameClient empty tag processing

tzonghao:
Empty or None tags should be left out, otherwise it would break line
protocol.


aviau:
Thank you for your work!

PR #371, opened 2016-09-22
Fix DataFrameClient import error on python3.5

vodik:
Was getting `NameError: name 'err' is not defined` on python3.5. Stashing the error message inside the class seems to fix the problem.


aviau:
Thanks!

PR #374, opened 2016-10-03
remove pytz requirement as not used anymore

bbinet:
pytz does not seem to be imported anywhere in the code.


agronholm:
It's used in line_protocol.py, line 12.

bbinet:
You're right.

PR #375, opened 2016-10-12
make_line as separate from make_lines

aisbaa:
Hi, extracted body of `make_lines` function to separate `make_line` function. Our input data has a bit different format than original function requires, therefore we couldn't use it. But it covers injection vulnerability.

Test cases fail because of pandas library deprecation warning.


aisbaa:
Any chance of merging this?

PR #386, opened 2016-11-28
Fix method name in example

NotSqrt:
A simple fix for a bug found by running the example.


aviau:
Thanks for the PR!

Thanks @xginn8 for reviewing!

Issue #381, opened 2016-10-25
InfluxDBClient.fromDSN should be classmethod instead of staticmethod


hvdklauw:
To add some logging for the queries being run I extended the InfluxDBClient class and overwrote the query method.

Only to have no queries being logged. After some digging I found that when I use the fromDSN method to create a client instance it's the original class, not my extended version.

Changing it to:

``` python
    @classmethod
    def from_DSN(cls, dsn, **kwargs):
        """Return an instance of :class:`~.InfluxDBClient` from the provided
        data source name. Supported schemes are "influxdb", "https+influxdb"
        and "udp+influxdb". Parameters for the :class:`~.InfluxDBClient`
        constructor may also be passed to this method.

        :param dsn: data source name
        :type dsn: string
        :param kwargs: additional parameters for `InfluxDBClient`
        :type kwargs: dict
        :raises ValueError: if the provided DSN has any unexpected values

        :Example:

        ::

            >> cli = InfluxDBClient.from_DSN('influxdb://username:password@\
localhost:8086/databasename', timeout=5)
            >> type(cli)
            <class 'influxdb.client.InfluxDBClient'>
            >> cli = InfluxDBClient.from_DSN('udp+influxdb://username:pass@\
localhost:8086/databasename', timeout=5, udp_port=159)
            >> print('{0._baseurl} - {0.use_udp} {0.udp_port}'.format(cli))
            http://localhost:8086 - True 159

        .. note:: parameters provided in `**kwargs` may override dsn parameters
        .. note:: when using "udp+influxdb" the specified port (if any) will
            be used for the TCP connection; specify the UDP port with the
            additional `udp_port` parameter (cf. examples).
        """

        init_args = parse_dsn(dsn)
        host, port = init_args.pop('hosts')[0]
        init_args['host'] = host
        init_args['port'] = port
        init_args.update(kwargs)

        return cls(**init_args)
```

Would allow the class to be extended.

Unless there is a specific reason for doing it this way?

xginn8:
this was fixed in 2e8c5886

PR #393, opened 2016-12-03
Fix for from_DSN() constructor when using DataFrameClient

eman:
When creating an instance of `DataFrameClient` using the `from_DSN()` constructor, an instance of `InfluxDBClient` is returned instead. This change returns the correct instance type.

See also issue #381


aviau:
Nice! Can we have a test for this?

aviau:
That is great, thanks!

PR #394, opened 2016-12-03
fixes for mutable default parameters in DataFrame client

eman:
There were a few places in the DataFrameClient where function definitions had mutable default parameters. This change address this and associated errors.


aviau:
Thanks!

PR #395, opened 2016-12-06
Use quote_ident() for database names in queries

cadnce:
Much like the other usages queries use the quote_ident() function for database names.

* Added an exception when the parameters will create a malformed request.
* Updated the tests around this to use a named database instead.

Not sure if raising an exception is the ideal thing to do here, thoughts?


aviau:
Raising an exception is perfectly fine. This will happen only when the client is used in a way that is not meant to be used.

aviau:
Thank you for your work.

PR #354, opened 2016-07-29
Remove cluster client.

gaker:
https://github.com/influxdata/influxdb-python/issues/339

Issue #339, opened 2016-06-13
Remove the cluster client


aviau:
The cluster client no longer makes sense with recent patches.

Don't forget:
- remove the example in the README.
- remove the section in api-documentation.rst

PierreF:
Hi,
which recent patches do you refer to ?
Or are you referring to InfluxDB itself that no longer support cluster ?

we are still using InfluxDB 0.11 with cluster, and that feature of influxdb-python was useful to us :(

aviau:
This refers to https://github.com/influxdata/influxdb-python/pull/354. It was removed because InfluxDB instances can no longer be considered as equivalent (in the open source version). You can't just write a point to another instance if one of them is down. And round robin does not make sense either.

https://github.com/influxdata/influxdb-relay Can come into help. One would use the python client to write to a relay instead of writing to any server in the pool.

PierreF:
Fine, thanks for taking time to explain the removal.

I've created a PR (#396) to re-add the cluster in a legacy module with warning that state this cluster client must NOT be used with newer influxdb-relay cluster.

Convolution:
We are using the latest commercial version of InfluxDB (v1.2 Enterprise) which does not rely on a relay server. This means the load balancing still needs to be done on the client, hence there is still a strong use case for the ClusterClient. The removal of this functionality means that we cannot upgrade to the latest release of influxdb-python. 

Are there any plans to support Influx-Enterprise?

PierreF:
PR #396 should works with latest InfluxDB enterprise cluster, it only require that all nodes are equivalent: writing on one node is replicated by the node to the other nodes; reading from any node return the same data

But since the ClusterClient is still useful in Influxdb enterprise, maybe the PR should not move it to a "legacy" module, but not normal module and/or "enterprise" module (e.g. with due warning that it does not works with relay cluster). What do you think @aviau, should I update the PR in this way ?

PR #396, opened 2016-12-09
Re-add cluster-client in legacy influxdb011 module

PierreF:
PR #354 removed the cluster support as requested in issue #339.
The removed is due to InfluxDB no longer support cluster with equivalent node (in open-source version).

But for users of older InfluxDB version (0.11) with old cluster, it still make sense. This PR re-add the cluster in a legacy "influxdb011" module.

This InfluxDBClusterClient must NOT be used with newer cluster (and README + pydoc of the module state this).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,6 +27,8 @@ InfluxDB pre v1.1.0 users
 
 InfluxDB 1.1.0 was released and it is the new recommended version. InfluxDB 0.8.x users may still use the legacy client by using ``from influxdb.influxdb08 import InfluxDBClient`` instead.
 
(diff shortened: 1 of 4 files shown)


PierreF:
Any update ? @aviau, since it still useful for InfluxDB enterprise cluster (cf https://github.com/influxdata/influxdb-python/issues/339#issuecomment-268482245) should I update the PR to re-add cluster in a non-legacy module (with warning that it don't work with open-source relay based cluster) ?

sebito91:
Closing due to staleness with enterprise cluster

PR #402, opened 2016-12-26
Add get_list_measurements and drop_measurement

Vic020:
Dear maintainer,

Look likes we are missing the get list measurements and drop measurement feature for client.

pls accept this pr.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -547,6 +547,32 @@ class InfluxDBClient(object):
         """
         self.query("DROP DATABASE {0}".format(quote_ident(dbname)))
 
+    def get_list_measurements(self):
+        """Get the list of measurements in InfluxDB.
+
+        :returns: all measurements in InfluxDB
+        :rtype: list of dictionaries
+
+        :Example:
+
+        ::
+
+            >> dbs = client.get_list_measurements()
+            >> dbs
+            [{u'name': u'measurements1'},
+             {u'name': u'measurements2'},
+             {u'name': u'measurements3'}]
+        """
+        return list(self.query("SHOW MEASUREMENTS").get_points())
+
+    def drop_measurement(self, measurementname):
+        """Drop a measurement from InfluxDB.
+
+        :param dbname: the name of the measurement to drop
+        :type dbname: str
+        """
+        self.query("DROP MEASUREMENT {0}".format(quote_ident(measurementname)))
+
     def create_retention_policy(self, name, duration, replication,
                                 database=None, default=False):
         """Create a retention policy for a database.
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -565,13 +565,13 @@ class InfluxDBClient(object):
         """
         return list(self.query("SHOW MEASUREMENTS").get_points())
 
-    def drop_measurement(self, measurementname):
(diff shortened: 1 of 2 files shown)


Vic020:
What does the tarvis-ci report mean?

doc building needs length check ?

pls, give me a prompt to fix this error

aviau:
The docs tests seems unrelated. I have restarted  the job.

Vic020:
@aviau Test case done.

but the doc building failed, again.

Vic020:
According to https://github.com/sphinx-doc/sphinx/issues/3212

sphinx 1.5.1 fix this doc building problem

influxdb/client.py
```diff
@@ -495,6 +495,32 @@ def drop_database(self, dbname):
...
+            [{u'name': u'measurements1'},
+             {u'name': u'measurements2'},
+             {u'name': u'measurements3'}]
+        """
+        return list(self.query("SHOW MEASUREMENTS").get_points())
+
+    def drop_measurement(self, measurementname):
+        """Drop a measurement from InfluxDB.
+
+        :param dbname: the name of the measurement to drop
```
smolse: Should be `measurementname` instead of `dbname` according to the method's signature. But even better would be to rename the parameter to `measurement` in order to be consistent with naming in other methods.

influxdb/client.py
```diff
@@ -495,6 +495,32 @@ def drop_database(self, dbname):
...
+
+        ::
+
+            >> dbs = client.get_list_measurements()
+            >> dbs
+            [{u'name': u'measurements1'},
+             {u'name': u'measurements2'},
+             {u'name': u'measurements3'}]
+        """
+        return list(self.query("SHOW MEASUREMENTS").get_points())
```
smolse: Would be good to add support for `on` clause which was added in InfluxDB 1.1.

Vic020: By checking  InfluxDB 1.2  docs about show measurements.
There was more than one clause needed to support. **on** , **with measurement**, **where**, **limit**, **offset**.
Support all the clauses maybe coming from another PR, and it's better that we could discuss the interface of this feature.

Vic020: @smolse 

By checking InfluxDB 1.2 docs about show measurements.
There was more than one clause needed to support. on , with measurement, where, limit, offset.
Support all the clauses maybe coming from another PR, and it's better that we could discuss the interface of this feature.

sebito91: I concur with that, there are features added regularly that we'll need to keep up with. Since we're officially supporting the v1.2.4 release this is fine as-is.

tox.ini
```diff
@@ -26,7 +26,7 @@ commands = nosetests -v --with-coverage --cover-html --cover-package=influxdb
 [testenv:docs]
 deps = -r{toxinidir}/requirements.txt
        pandas
-       Sphinx==1.2.3
+       Sphinx==1.5.1
```
sebito91: Can you please remove this, we're already up to v1.5.5?

Vic020:
@sebito91 Thanks. 

You are right, already deleted Sphinx version modification, and added docstring for tests.

Hope to receive your feedback.

PR #404, opened 2017-01-02
Space padded select-queries should return DataFrames

AndreCAndersen:
Fixed an issue where space padded select-queries are interpreted as being non-select-queries thus falling back to returning a ResultSet instead of a dictionary of panda DataFrames.


AndreCAndersen:
This pull request should address [issue 425](https://github.com/influxdata/influxdb-python/issues/425).

aviau:
Thanks!

PR #407, opened 2017-01-17
Fixing rounding errors when inserting microseconds

AndreCAndersen:
When calculating nanoseconds since EPOCH, `total_seconds()` will give a float representation which when multiplied by 1e9 and cast to `int` will create rounding errors. E.g. inserting `2017-01-17T08:00:12.000001` with a datetime will produce `2017-01-17T08:00:12.000001024Z` in InfluxDB. This commit fixes this by never going via floating seconds, and is based on the `total_seconds()` function itself.


influxdb/line_protocol.py
```diff
@@ -16,6 +16,12 @@
 EPOCH = UTC.localize(datetime.utcfromtimestamp(0))
 
 
+def _to_nanos(timestamp):
+    delta = timestamp - EPOCH
+    nanos = (delta.days * 86400 + delta.seconds) * 10 ** 9 + delta.microseconds * 10 ** 3
```
xginn8: this line fails flake8 (89 chars), mind splitting it up? otherwise, this looks like a good change.

AndreCAndersen: [How about now](https://github.com/influxdata/influxdb-python/pull/407/commits/01bc900a9e2f4dde1471f56cdcf34a9394e04d1e)?

clslgrnc:
Just encountered this rounding error and was about to submit this very patch.
Can this be merged?

In order to ensure that ns stays an int I would also rewrite:
```python
        if precision is None or precision == 'n':
            return ns
        elif precision == 'u':
            return ns / 1e3
        elif precision == 'ms':
            return ns / 1e6
        elif precision == 's':
            return ns / 1e9
        elif precision == 'm':
            return ns / 1e9 / 60
        elif precision == 'h':
            return ns / 1e9 / 3600
```

to
```python
        if precision is None or precision == 'n':
            return ns
        elif precision == 'u':
            return ns / 10**3
        elif precision == 'ms':
            return ns / 10**6
        elif precision == 's':
            return ns / 10**9
        elif precision == 'm':
            return ns / 10**9 / 60
        elif precision == 'h':
            return ns / 10**9 / 3600
```

clslgrnc:
I would probably allow you to close some of those:
https://github.com/influxdata/influxdb-python/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+nanoseconds

sebito91:
Duplicate of #649 and fixed in #650

sebito91:
Re-addressed in a new PR that merges this content including fixes from @clslgrnc

PR #408, opened 2017-01-20
Changes batching to allow all iterables

nomad010:
Not all iterables support len or indexing. This changes the implementation of the batching function so that it supports all iterables. This is important to support large ingestion sizes.


nomad010:
Not sure why the tox test failed, so I cannot fix if it is the code.

pkittenis:
Tests should pass now once the branch is rebased on latest master.

xginn8:
Superseded by #746

PR #409, opened 2017-01-23
add ping method to the client

pmenglund:
a method to invoke the `/ping` endpoint so it is easy to verify connectivity


influxdb/client.py
```diff
@@ -410,6 +410,18 @@ def write_points(self,
                                       retention_policy=retention_policy,
                                       tags=tags, protocol=protocol)
 
+    def ping(self):
+        """Check connectivity to influxdb
```
smolse: Minor thing, but summary should end in a period and `influxdb` replaced with `InfluxDB`, so that it will be in the same style as other docstrings. Also, there should be a blank line after the summary line (PEP 257). Otherwise, having this method would be really useful in my opinion 👍

pmenglund: Sure, I'll fix that

influxdb/client.py
```diff
@@ -411,9 +411,10 @@ def write_points(self,
                                       tags=tags, protocol=protocol)
 
     def ping(self):
-        """Check connectivity to influxdb
-        :returns: The version of the influxdb the client is connected to
+        """Check connectivity to InfluxDB
+        :returns: The version of the InfluxDB the client is connected to
         """
+
```
smolse: Blank line should be after the summary line (i.e. the first line of the docstring), before the description. In PEP 257 no blank lines allowed after the method's docstring. Also, please note that PEP 257 requires that the first line of the docstring ends with the period. 
On the other side, there are currently quite many inconsistencies in the client's docstrings and tox does not check code for PEP 257. May be all these inconsistencies should be revised in a separate PR.

pmenglund: Doh! I'll fix that, and create an issue for fixing pep257 in the remaining files

pmenglund: Created #415

PR #414, opened 2017-02-11
Add CQs management methods to the client

smolse:
This PR adds to the client new methods which allow create, show and drop continuous queries in InfluxDB more easily.


smolse:
@aviau seems like docs test is really broken.

aviau:
>  seems like docs test is really broken.

Yeah. Somebody needs to fix it.

aviau:
Can you rebase this on master? The tests should now pass.

smolse:
Yay, it is passing now.

sebito91:
@smolse the logic here is completely fine, but can you please add comments to the test functions (something basic)? We've introduced PEP257 compliance over the past few weeks, and that's the only failing test at the moment.

Once you've added, we can merge this.

lukaszdudek-silvair:
@smolse - do you intend to work on this anytime soon?
AFAIU the thing to be done (apart from rebasing on `master` again) is to add docstrings to test cases - is that right, @sebito91 ?

PR #418, opened 2017-02-16
[Ready for Review] Chunked query responses

pkittenis:
Changes:

* Added `chunked` and `chunk_size` keyword arguments to InfluxDBClient.query
* Added chunked response parsing of queries and chunked response test
* Added allow failure on docs for travis-ci

`query` doc strings now indicate that with `chunked=True` a _generator_ of `ResultSet` objects is returned as opposed to a list.

On latest master branch trying to use chunked responses by setting `params={'chunked': 'true'}` would result in JSON parse failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ addons:
     - wget
 
 matrix:
+  allow_failures:
+    - python: 3.4
+      env: TOX_ENV=docs
   include:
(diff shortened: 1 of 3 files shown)


aviau:
Thanks for the good PR!

pkittenis:
My pleasure :smile:

Issue #400, opened 2016-12-20
I found a bug in influxdb-python


JasonLZJ:
I use python2.6 . When I use pip install influxdb , influxdb-python 4.0.0 will be setup.
But when I save data , It will touch off an error looks like : AttributeError: 'datetime.timedelta' object has no attribute 'total_seconds'
  I uninstall it and install influxdb-python 3.0.0 ,everyting is ok.
  I think the cause of the issue is "datetime.timedelta.total_seconds" was added in python2.7
  So I create this issue.

aviau:
We won't fix it, because we deprecated python2.6 support. Sorry.

However, feel free to open a PR and we will merge it.

PR #428, opened 2017-03-17
Added 'total_seconds' function for python26 compatibility

mmitucha:
Related to #400 
Maybe not so elegant solution, but have been forced to make it work with Python2.6 project.
It would be very nice if you could merge this or make it work with python2.6 somehow.
Thank you

### DIFF
--- a/influxdb/influxdb08/dataframe_client.py
+++ b/influxdb/influxdb08/dataframe_client.py
@@ -13,16 +13,6 @@ import warnings
 from .client import InfluxDBClient
 
 
-def total_seconds(td):
-    # Keep backward compatibility with Python 2.6 which doesn't have
-    # this method
-    if hasattr(td, 'total_seconds'):
-        return td.total_seconds()
-    else:
-        return (td.microseconds +
-                (td.seconds + td.days * 24 * 3600) * 10**6) / 10**6
-
-
 class DataFrameClient(InfluxDBClient):
     """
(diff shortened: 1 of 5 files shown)


influxdb/influxdb08/dataframe_client.py
```diff
@@ -13,6 +13,16 @@
 from .client import InfluxDBClient
 
 
+def total_seconds(td):
```
xginn8: Rather than duplicating this function, can you add it as a method to the InfluxDBClient class?

mmitucha: Hi, 
thank you for your time.
I tried to add `total_seconds` method to `InfluxDBClient` class, t will be more elegant solution, but problem is that `InfluxDBClient` class is not imported in **line_protocol.py**, and can't be because of "circular import error". 
So the `total_seconds` function must be separately specified in **line_protocol.py**.

See in new commit.

mmitucha: I do understand that this workaround is uglyfying the code for unsupported version of Python. Maybe we could put it into separate branch or I can leave it just in my fork.

aviau: But it could be moved to influxdb/total_seconds.py?

This does not need to be a class method.

mmitucha:
Terribly sorry for reposnse. Hopefully no one need this any more. So I will close this. If I am wrong, please let me know.

PR #429, opened 2017-03-21
Add correct protocol to the Pandas client

jackzampolin:
@nicolajkirchhof Is this what you were talking about?


aviau:
I don't really keep the dataframe example up to date, I'll take your word on that.

Issue #415, opened 2017-02-11
pep257 compliance


pmenglund:
Make the repo [pep257](https://www.python.org/dev/peps/pep-0257/) compliant.

There are currently 484 errors 😱

sebito91:
That's fair, will work on this and let you know when it's ready. PEP is HARD! 👊

aviau:
If we do this, it would be great to add a linter to travis.

sebito91:
That's a really good call, will look into that as well.

PR #430, opened 2017-03-22
fix for lost precision on float field values

isilien:
use `repr()` instead of `str()` when casting floats into strings so that `float(repr(my_float)) == my_float` can be guaranteed, which is not the case when using `str()`

### DIFF
--- a/influxdb/tests/test_line_protocol.py
+++ b/influxdb/tests/test_line_protocol.py
@@ -119,3 +119,19 @@ class TestLineProtocol(unittest.TestCase):
             line_protocol.quote_literal(r"""\foo ' bar " Örf"""),
             r"""'\\foo \' bar " Örf'"""
         )
+
+    def test_float_with_long_decimal_fraction(self):
+        data = {
+            "points": [
+                {
+                    "measurement": "test",
+                    "fields": {
+                        "float_val": 1.0000000000000009,
+                    }
+                }
+            ]
+        }
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


influxdb/tests/test_line_protocol.py
```diff
@@ -119,3 +119,19 @@ def test_quote_literal(self):
             line_protocol.quote_literal(r"""\foo ' bar " Örf"""),
             r"""'\\foo \' bar " Örf'"""
         )
+
+    def test_float_with_long_decimal_fraction(self):
```
xginn8: To conform to pep257 this public method should have a docstring, see #415

isilien: Fixed.

aviau:
thanks to you both!

chripede:
Any chance this could cause Decimal values to fail?

`InfluxDBClientError: 400: {"error":"unable to parse 'sale,host=app,payment_type=downgrade value=Decimal('29.95') 1492496475567468032': invalid boolean"}`

I think the problem might be that float() will return a float on a Decimal, but repr will actually return `"Decimal('29.95')"`

chaica:
I guess I do hae something like this here ::
```
influxdb.exceptions.InfluxDBClientError: 400: {"error":"partial write: unable to parse 'totalgain value=Decimal('2695.30')': invalid boolean dropped=0"}
```
Is there a workaround?

PR #433, opened 2017-03-28
improved make_lines efficiency

RonRothman:
This PR reduces some unneeded allocations and operations in `make_lines`.

We're using the influxdb client in our high throughput, low latency web service. We've profiled our service and found that `make_lines` was a bottleneck. When `make_lines` takes a long time, it blocks other greenlets from servicing requests.

No change in functionality; only efficiency.

### DIFF
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -5,13 +5,12 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from copy import copy
 from datetime import datetime
 from numbers import Integral
 
 from pytz import UTC
 from dateutil.parser import parse
-from six import binary_type, text_type, integer_types, PY2
+from six import iteritems, binary_type, text_type, integer_types, PY2
 
 EPOCH = UTC.localize(datetime.utcfromtimestamp(0))
 
@@ -108,7 +107,7 @@ def make_lines(data, precision=None):
     matching the line protocol introduced in InfluxDB 0.9.0.
     """
     lines = []
-    static_tags = data.get('tags', None)
+    static_tags = data.get('tags')
     for point in data['points']:
         elements = []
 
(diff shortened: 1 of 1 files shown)


influxdb/line_protocol.py
```diff
@@ -119,32 +118,29 @@ def make_lines(data, precision=None):
...
         # tags should be sorted client-side to take load off server
-        for tag_key in sorted(tags.keys()):
+        for tag_key, tag_value in sorted(iteritems(tags)):
             key = _escape_tag(tag_key)
-            value = _escape_tag(tags[tag_key])
+            value = _escape_tag(tag_value)
 
             if key != '' and value != '':
-                key_values.append("{key}={value}".format(key=key, value=value))
+                key_values.append(key + "=" + value)
```
xginn8: I believe it's better to use string formatting methods rather than appending strings via the `+` operator, can you please restore that behavior?

influxdb/line_protocol.py
```diff
@@ -119,32 +118,29 @@ def make_lines(data, precision=None):
...
+        for field_key, field_value in sorted(iteritems(point['fields'])):
             key = _escape_tag(field_key)
-            value = _escape_value(point['fields'][field_key])
+            value = _escape_value(field_value)
             if key != '' and value != '':
-                field_values.append("{key}={value}".format(
-                    key=key,
-                    value=value
-                ))
+                field_values.append(key + "=" + value)
```
xginn8: ditto above regarding concatenating strings

influxdb/line_protocol.py
RonRothman: @xginn8 Thanks for the review!

Better in what way?  `+` is generally one of the fastest methods of concatenating short strings; conversely, `format` with keywords is one of the slowest.

Because `key + "=" + value` is short and readable, and because this code is in a frequently executed loop body, I saw no need to sacrifice performance--we can have both efficient and readable code.  Thoughts?

xginn8: I stand corrected, that's my fault for not running some tests 👍. Looks like a good change.

pkittenis: `key_values.append("=".join([key, value]))` would be faster actually.

String concatenation creates temporary strings for each concatenation and will get slower as strings gets larger. 

`join` OTOH is implemented in C and does the whole operation in one go.

pkittenis: Speed wise would actually be best to replace the whole loop with a list comprehension:

```
key_values = ["=".join((key, _escape_tag(val))) for key, val in sorted(iteritems(tags))]
```

Cleaner and more readable to boot.

aviau:
Thanks!

PR #435, opened 2017-04-10
Allow user-configurable number of retries when submitting request to …

xginn8:
…database


aviau:
Nice! Can you add a test for that? There is already this one:
 - https://github.com/influxdata/influxdb-python/blob/master/influxdb/tests/client_test.py#L625

Maybe add another one that configures the retry count to something else.

aviau:
This looks good to me :)

PR #436, opened 2017-04-10
Allow np.nan, np.inf values in DataFrameClient.write_points

### LINKED ISSUE #422: DataFrame write "nan", "inf" error in influxdb
see here https://github.com/influxdata/influxdb-python/issues/195
It seems that in the latest version, if a DataFrame contains nan, inf elements, there will be error when writing to influxdb. The problem is that the line protocol doesn't support "value1=nan" or "value1=inf". You can drop rows containing nan using df.dropna(), but in most of the case, in the same row, you will need to keep those values that are not "nan", so droping an entire row is not ideal. The solution can be to omit the value fields that is "nan" and "inf" (np.isfinite()), something like, changing "value1=0.3, value2=nan, value3=inf"  to "value1=0.3". In this way, the influxdb will keep value1, and give value2, and value3 empty values.

nmerket:
Fixes #422


.gitignore
```diff
@@ -21,3 +21,4 @@ docs/build/
 .coverage
 cover
 env
+.idea
```
aviau: This isn't produced by anything in our repository. It should be ignored globally on your system.

See: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

nmerket:
Fair enough. I can remove that.


On Apr 11, 2017, at 7:42 PM, Alexandre Viau <notifications@github.com<mailto:notifications@github.com>> wrote:


@aviau commented on this pull request.

________________________________

In .gitignore<https://github.com/influxdata/influxdb-python/pull/436#discussion_r111050479>:

> @@ -21,3 +21,4 @@ docs/build/
 .coverage
 cover
 env
+.idea


This isn't produced by anything in our repository. It should be ignored globally on your system.

See: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

-

jacobeverist:
Thanks for this!  Missing data is all over my datasets and it made the pandas influxdb interface completely unusable without this patch.

nmerket:
@jacobeverist You're welcome. I made this quite a while ago to solve a problem I was having and then didn't need to use influxdb after all for my project. I'm not sure why it was never merged. @aviau had one minor comment, and I addressed it. It looks like there's another PR that addresses this that's pending as well (#507).

sebito91:
Replaced by #816, yay!

PR #439, opened 2017-04-26
Add support for DataFrameClient chunked queries

gusutabopb:
`DataFrameClient` currently does not properly handle chunked requests [[StackOverflow question](http://stackoverflow.com/questions/42603799/how-to-retrive-more-than-10k-lines-from-influxdb-using-pandas/)]. This PR fixes that.


influxdb/_dataframe_client.py
```diff
@@ -131,25 +132,26 @@ def write_points(self,
...
 
-        :param chunked: [Optional, default=False] True if the data shall be
-            retrieved in chunks, False otherwise.
+        :param query: the actual query string
+        :param dropna: drop columns where all values are missing
+        :param **kwargs: additional parameters for ``InfluxDBClient.query``
 
         """
-        results = super(DataFrameClient, self).query(query, database=database)
+        results = super(DataFrameClient, self).query(query, **kwargs)
```
pkittenis: Best to use named keyword arguments rather than `kwargs` magic. Any arguments passed to dataframe client not present in the super class function will cause a run time stack trace.

gusutabopb: The rationale behind using `kwargs` is just to make the code more compact and maintainable.

 `DataFrameClient.query` is nothing more than a wrapper around `InfluxDBClient.query`. I personally think wrappers should be as thin as possible and do not see the need to be more verbose and copy-and-paste all the extra optional arguments. Adding all the optional arguments of `InfluxDBClient.query` to `DataFrameClient.query` will also make future maintenance more difficult and error prone. Whenever there is some change in the signature of `InfluxDBClient.query`, that change will need to be manually reflected to all wrapper functions.

This `kwargs` approach is widely used in libraries that have a lot of functions wrapping other functions with a large number of arguments. Matplotlib is the best example that comes up to my mind.

As for passing inexistent arguments causing a runtime error, well, the same can be said for passing inexistent arguments directly to `InfluxDBClient.query`. The only REAL disadvantage of the `kwargs` argument I can think of is the lack of IDE tab completion help. 

So, IMHO, it boils down to a trade off between maintainability+compactness vs IDE completion. I personally think the former to be way more important, but I will commit some code reflecting your change request.

pkittenis: | As for passing inexistent arguments causing a runtime error well, the same can be said for passing inexistent arguments directly to InfluxDBClient.query

The difference though is, without `kwargs` you get warnings from static analyser tools like pylint, flake8 etc, while with `kwargs` code will only break when it tries to execute at run time. That would be fine for internal functions which are only used within library but this is a public endpoint.

Taking that argument to its extreme, why even use named keyword parameters in the first place? Just use `kwargs` everywhere. You could, usage wise will work the same way, but the code becomes unreadable and users don't know what arguments they can pass in without reading documentation or going through code.

It's more code to write and maintain sure, but is better practice and clearer from a user's point of view, meaning looking at method signature tells you its parameters. 

As always though, these things are subject to opinion - you may not agree and that's fine! :) It's really up to maintainers to set guidelines on things like this.

gusutabopb: Thanks for the feedback. 

As you said, "maintainers to set guidelines on things like this", so I will just follow what the maintainers decide then :)

I hadn't really thought about static analyser tools and about the relation between `kwargs` and private/public functions. I am glad we had this little discussion.

Have you taken a chance to look at my latest commits? Please let me know if you think the PR is mergeable now. If not, please let me know what else I should change.

sebito91:
@gusutabopb this looks solid, well done! Could you please rebase + squash this request and we'll look to merge?

@aviau ready to merge when squashed!

influxdb/_dataframe_client.py
pkittenis: +1 from me :)

aviau:
> Can you please rebase and squash your commits, then will merge.

You can do that yourself with the "squash and merge" button:

screenshot from 2017-08-15 20-20-56

gusutabopb:
@sebito91 Sorry for the late response. This is my first PR and I couldn't find that squash & merge button:


screen shot 2017-08-16 at 6 00 15



It seems it's merged now (thanks!), but the git history is kind of dirty with all the intermediate commits I made to update the branch to master. Is there anything I can do to fix that (or at least avoid it going forward)? Thanks!

sebito91:
That is correct, you won't see the options for merging because you don't have write permission to the repo. It's not ideal for us to force it through the GUI, but hey it works. In the future, you would do something as follows:

```
1. git pull --verbose --rebase upstream master
2. count the number of commits between previous HEAD and your new HEAD (in this case it looks like 9)
3. git rebase --interactive --verbose HEAD~9
4. leave the first line as pick, mark the other 8 as fixup
5. save and quit that file
6. git push --verbose --force origin <branchname>
```

Good practice when submitting PRs is to either mark subsequent commits as fixup commits (https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) which can help you do that rebase I mentioned or just squash them each time when you submit changes based on feedback or requests.

We'll fix things up for the next time, but thank you very much for your effort here! If you have any more questions please feel free to email me or hop onto #influxdb in freenode.

PR #440, opened 2017-04-26
fix the bug of function get_points when with tags parameters

nuaays:
when i used influxdb-python to query with tags as follows, the result showed nothing
```
result = client.query('select * from cpu limit 100;')
for p in list(result.get_points( measurement='cpu', tags={"host":"10.12.0.99"} )):
      print p
```

so, i checked the function of get_points and the function _tag_matches, then i found two errors in get_points
 
* 1.change serie.get('`tags`', {}) to serie.get('`columns`', {}) 
* 2.move condition judgement line which invoke _tag_matches  into for loop, and change `serie_tags` to `item`
```
 -                serie_tags = serie.get('tags', {})
 -                if tags is None or self._tag_matches(serie_tags, tags):
 -                    for item in self._get_points_for_serie(serie):
 +                serie_tags = serie.get('columns', {})
 +                for item in self._get_points_for_serie(serie):
 +                    if tags is None or self._tag_matches(item, tags):   
```

### DIFF
--- a/influxdb/resultset.py
+++ b/influxdb/resultset.py
@@ -106,7 +106,7 @@ class ResultSet(object):
                 # we will matches every returned serie
                 serie_tags = serie.get('columns', {})
                 for item in self._get_points_for_serie(serie):
-                    if tags is None or self._tag_matches(item, tags):   
+                    if tags is None or self._tag_matches(item, tags):
                         yield item
 
     def __repr__(self):
--- a/influxdb/resultset.py
+++ b/influxdb/resultset.py
@@ -104,9 +104,9 @@ class ResultSet(object):
             elif measurement in (None, serie_name):
                 # by default if no tags was provided then
                 # we will matches every returned serie
-                serie_tags = serie.get('tags', {})
-                if tags is None or self._tag_matches(serie_tags, tags):
-                    for item in self._get_points_for_serie(serie):
+                serie_tags = serie.get('columns', {})
+                for item in self._get_points_for_serie(serie):
+                    if tags is None or self._tag_matches(item, tags):   
                         yield item
 
     def __repr__(self):


nuaays:
the ci failed, but i checked the sample real cpu data in influxdb, different data json structure which is the root cause of "test_filter_by_tags" failed , i think

```json
{u'series': [{u'columns': [u'time',
                           u'cpu',
                           u'host',
                           u'usage_guest',
                           u'usage_guest_nice',
                           u'usage_idle',
                           u'usage_iowait',
                           u'usage_irq',
                           u'usage_nice',
                           u'usage_softirq',
                           u'usage_steal',
                           u'usage_system',
                           u'usage_user'],
              u'name': u'cpu',
              u'values': [[u'2017-04-26T03:56:10Z',
                           u'cpu-total',
                           u'10.12.0.100',
                           0,
                           0,
                           89.32726352916748,
                           0.05058168942829195,
                           0,
                           0,
                           0,
                           0,
                           3.5407182598884286,
                           7.081436519776857],
                          [u'2017-04-26T03:56:10Z',
                           u'cpu1',
                           u'10.12.0.99',
                           0,
                           0,
                           93.64919354946599,
                           0.10080645161356867,
                           0,
                           0,
                           0,
                           0,
                           2.4193548387439847,
                           3.830645161366035]]}],
 u'statement_id': 0}
```

nuaays:
@voidfiles

aviau:
Did the response format change? Can you please update the tests?

aviau:
It looks like columns is a new concept in InfluxDB, but tags still exist?

sebito91:
I have to defer to @aviau to confirm either way, but...

To be fair, the query you request is open-ended and effectively returns the raw data free from filtering and/or grouping. Though this is fine when you want to get at the raw data, the structure of the response doesn't present any tags for you to filter by unless you define that in your query directly. In other words, it's expected behavior to see just the columns returned with their respective values.

The function you're trying to leverage requires some kind of grouping to present tags that can be accessed. To do what you're after _without_ amending existing code, you'd simply need to update your query to:

```select * from cpu group by host limit 100```

Do you want to give that a shot and let us know? If that doesn't work, we can discuss further adding your PR to the codebase.

sebito91:
@nuaays have you had a chance to check out the above? ^^

sebito91:
Closing stale issue, work already incorporated into master.

PR #442, opened 2017-04-26
Added support for python 2.6.6

eorochena:
Added work around for AttributeError: 'datetime.timedelta' object has no attribute 'total_seconds' as suggested in  [https://docs.python.org/2/library/datetime.html#datetime.timedelta.total_seconds](url).

Even though Python 2.6.6 has been EOL for quite a while, there are still some systems that have it as their main Python interpreter like CentOS 6 and the only way to use a more recent version of Python is to install the SCL repo along with a few additional changes which a little bit more complicated or not feasible if the user doesn't have admin rights.

### DIFF
--- a/influxdb/__init__.py
+++ b/influxdb/__init__.py
@@ -9,7 +9,6 @@ from .client import InfluxDBClient
 from .dataframe_client import DataFrameClient
 from .helper import SeriesHelper
 
-
 __all__ = [
     'InfluxDBClient',
     'DataFrameClient',
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -23,19 +23,21 @@ def _convert_timestamp(timestamp, precision=None):
     if isinstance(timestamp, datetime):
         if not timestamp.tzinfo:
             timestamp = UTC.localize(timestamp)
-        ns = (timestamp - EPOCH).total_seconds() * 1e9
-        if precision is None or precision == 'n':
-            return ns
-        elif precision == 'u':
-            return ns / 1e3
-        elif precision == 'ms':
-            return ns / 1e6
-        elif precision == 's':
-            return ns / 1e9
-        elif precision == 'm':
-            return ns / 1e9 / 60
-        elif precision == 'h':
-            return ns / 1e9 / 3600
+            td = (timestamp - EPOCH)
+            ns = ((td.microseconds + (td.seconds + td.days * 24 * 3600) *
+                   10 ** 6) / 10 ** 6) * 1e9
+            if precision is None or precision == 'n':
+                return ns
+            elif precision == 'u':
+                return ns / 1e3
+            elif precision == 'ms':
+                return ns / 1e6
+            elif precision == 's':
+                return ns / 1e9
+            elif precision == 'm':
+                return ns / 1e9 / 60
+            elif precision == 'h':
+                return ns / 1e9 / 3600
     raise ValueError(timestamp)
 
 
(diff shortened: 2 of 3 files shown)


aviau:
If you want, you could re-enable python2.6 tests on travis and that would help us spot mistakes like these. But you would need to make sure that all tests pass.

Also, can you please fix flake8?

eorochena:
fixed previous pep8 issues, I'm able to run the code on Python 2.6 without issues but travis-ci fails with AttributeError: 'module' object has no attribute 'skipIf', [https://travis-ci.org/influxdata/influxdb-python/jobs/226203151](url), it doesn't appear to be a show stopper but I will need to investigate this further

influxdb/__init__.py
```diff
@@ -9,7 +9,6 @@
 from .dataframe_client import DataFrameClient
 from .helper import SeriesHelper
 
-
```
xginn8: was this deleted accidentally?

influxdb/line_protocol.py
```diff
@@ -23,19 +23,36 @@ def _convert_timestamp(timestamp, precision=None):
...
+                return ns / 1e9
+            elif precision == 'm':
+                return ns / 1e9 / 60
+            elif precision == 'h':
+                return ns / 1e9 / 3600
+        except AttributeError:
+            td = (timestamp - EPOCH)
+            ns = ((td.microseconds + (td.seconds + td.days * 24 * 3600) *
+                   10 ** 6) / 10 ** 6) * 1e9
+            if precision is None or precision == 'n':
```
xginn8: this section can be removed from the try/except, as it is common to both cases once you've defined `ns`

influxdb/__init__.py
eorochena: I just did a clone from the original repository and the file is still the same 

```
eorochena@Alchemy ~/GitHub/placeholdr/influxdb-python/influxdb $ git remote show origin
* remote origin
  Fetch URL: https://github.com/influxdata/influxdb-python.git
  Push  URL: https://github.com/influxdata/influxdb-python.git
  HEAD branch: master
  Remote branch:
    master tracked
  Local branch configured for 'git pull':
    master merges with remote master
  Local ref configured for 'git push':
    master pushes to master (up to date)

eorochena@Alchemy ~/GitHub/placeholdr/influxdb-python/influxdb $ diff --side-by-side __init__.py ../../influxdb-python/influxdb/__init__.py 
# -*- coding: utf-8 -*-						# -*- coding: utf-8 -*-

from __future__ import absolute_import				from __future__ import absolute_import
from __future__ import division					from __future__ import division
from __future__ import print_function				from __future__ import print_function
from __future__ import unicode_literals				from __future__ import unicode_literals

from .client import InfluxDBClient				from .client import InfluxDBClient
from .dataframe_client import DataFrameClient			from .dataframe_client import DataFrameClient
from .helper import SeriesHelper				from .helper import SeriesHelper


__all__ = [							__all__ = [
    'InfluxDBClient',						    'InfluxDBClient',
    'DataFrameClient',						    'DataFrameClient',
    'SeriesHelper',						    'SeriesHelper',
]								]


__version__ = '4.1.0'						__version__ = '4.1.0'
eorochena@Alchemy ~/GitHub/placeholdr/influxdb-python/influxdb $ md5sum __init__.py ../../influxdb-python/influxdb/__init__.py 
b8d0b62421afe8e07204abd8da548e80  __init__.py
b8d0b62421afe8e07204abd8da548e80  ../../influxdb-python/influxdb/__init__.py
```

influxdb/line_protocol.py
eorochena: @xginn8 Thanks for pointing this out, I just implemented the suggested changes

PR #443, opened 2017-05-02
added support for python 2.6.6

eorochena:
fixing flake8/pep8 issues

fixing flake8/pep8 issues again

travis for Python 2.6

fixing AttributeError on python 2.6.9

pep8

pep8

travis checks for 2.6.x

unittest for travis checks for 2.6.x

passing travis tests 2.7 - forward

implemented changes suggested by @xginn8

copied requirements and __init__.py from upstream

Added back the AttributeError exception

Added back the AttributeError exception and included datetime

removed comment


eorochena:
Corrected issues with previous pull request.

aviau:
Can we add 2.6 tests on travis?

sebito91:
Can you please rebase against existing master and ensure this check passes the new set of tests?

@aviau UPDATE REQUESTED

sebito91:
@eorochena, any chance you'd be able to rebase + add a test or two for py2.6.6? We can add this into the travis CI

eorochena:
Sorry, I was a little busy, I'll see what I can do to make this happen.

eorochena:
finally updated branch, sorry for the delay

sebito91:
That's fine, thanks for updating. Would you mind please adding a test case to your PR for python v2.6.6 to ensure everything passes?

eorochena:
For some reason Travis CI keeps failing with 

> ERROR: InvocationError: '/home/travis/build/influxdata/influxdb-python/.tox/py27/bin/nosetests -v --with-doctest'

I can't pinpoint where the actual error is

influxdb/tests/py26_test.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python
```
sebito91: You should change this to `#!/usr/bin/env python` otherwise this might not use the correct version.

influxdb/tests/py26_test.py
```diff
@@ -0,0 +1,57 @@
...
+import requests
+import time
+
+from influxdb import InfluxDBClient
+
+# InfluxDB Server parameters
+influxdb_server = "localhost"
+port = 8086
+database_name = "test_db"
+ip_addr = os.popen("hostname").read()
```
sebito91: This is somewhat problematic, you should use the socket infra instead of `popen`.

```
In [1]: import socket

In [2]: ip = socket.gethostbyname(socket.gethostname())

In [3]: ip
Out[3]: '192.168.1.77'
```

influxdb/tests/py26_test.py
```diff
@@ -0,0 +1,57 @@
...
+client = InfluxDBClient(influxdb_server, port, "influx-user", "influx-pass")
+
+try:
+    client.create_database(database_name)
+except requests.ConnectionError as e:
+    print("Failed to start, unable to establish connection with InfluxDB server %s - %s" % (influxdb_server, e))
+    sys.exit(2)
+
+# System Load
+load_avg_1 = os.popen("cat /proc/loadavg|awk '{print $1,$2,$3}'").read()[0]
```
sebito91: Can you please look to use the `subprocess.check_output` method here? The `os.popen` commands have been deprecated.

https://docs.python.org/2/library/subprocess.html#subprocess.check_output
https://docs.python.org/2/library/subprocess.html#replacing-os-popen-os-popen2-os-popen3

influxdb/tests/py26_test.py
```diff
@@ -0,0 +1,57 @@
...
+    client.write_points(load_5)
+
+    load_15 = [
+        {"measurement": "Load_15_minute",
+         "tags": {"hosts": ip_addr},
+         "time": datetime.datetime.now(), "fields": {"load_avg_15": int(load_avg_15)}}]
+    client.write_points(load_15)
+
+    counter += 1
+    time.sleep(3)
```
sebito91: This is a long sleep, mind making this 1s?

influxdb/line_protocol.py
```diff
@@ -40,8 +42,7 @@ def _convert_timestamp(timestamp, precision=None):
             return ns / 1e9 / 60
         elif precision == 'h':
             return ns / 1e9 / 3600
-
-    raise ValueError(timestamp)
+        raise ValueError(timestamp)
```
sebito91: This should be outdented, otherwise what happens if we don't hit any of the first two `isinstance` checks?

influxdb/tests/py26_test.py
eorochena: I will use socket.gethostname() because socket.gethostbyname(socket.gethostname()) tends to give the loopback interface ip address this value is pulled from the /etc/hosts file by default on stuck configuration 127.0.0.1 is associated to the hostname

influxdb/tests/py26_test.py
eorochena: unfortunately, check_output is not available on python2.6

influxdb/tests/py26_test.py
eorochena: sure, no problem

eorochena:
Implemented requested changes

sebito91:
Was thinking to add something like [this](https://github.com/influxdata/influxdb-python/blob/master/influxdb/tests/server_tests/influxdb_instance.py#L21) for `check_output` support in the test, but to be fair this should be refactored to handle more cleanly.

sebito91:
The issue is in your py26_test.py, which makes sense since it's not using proper `unittest.TestCase` format like other tests. I'll submit a refactored test for you shortly, stay tuned.

sebito91:
OK, if you move remove your current `py26_test.py` and create a new test (you can call it the same thing) under `~/influxdb/tests/server_tests/<filename>.py`, then replace your contents with the following from [here](https://pastebin.com/BCg9yBNm) you should be able to pass all of the tests.

We'll need to refactor some of the test code to bring these helpers out into the harness more easily, but for now this should get you working.

eorochena:
Implemented requested changes all travis ci checks passed

sebito91:
We're not supporting <python2.7 at this point. Closing issue...

PR #453, opened 2017-05-18
Update client.py to support InfluxDB access over reverse proxy

GilZm:
Add 'uri' optional parameter to be able to use reverse proxy !

Can also be done by overloading the existing InfluxDBClient object :
 `class InfluxDBClient_RP(InfluxDBClient):

         def __init__(self,**kwarg):
             self.__uri= kwarg.get('uri','')
 
             if kwarg.has_key('uri'):
                 del(kwarg['path'])
             super(InfluxDBClient_RP, self).__init__(**kwarg)
 
         def request(self, url, **kwarg):
             super(InfluxDBClient_RP, self).request("{0}/{1}".format(self.__uri, url), **kwarg)`

then configure for example apache like this  :
` <Proxy http://localhost:8086/*>
      Allow from all
 </Proxy>
 <LocationMatch "/influxproxy">
      ProxyPass http://localhost:8086 retry=0
      ProxyPassReverse http://localhost:8086
      RequestHeader set X-Forwarded-For %{REMOTE_ADDR}s
 </LocationMatch>`

finally use the uri param while creating the client :

 `client = InfluxDBClient_RP(host='myserver.domain.com', uri='influxproxy', port=443, ssl=True, database='somedatabase')
client.write_points(points)`

> 'insert code' is not working fine !?


aviau:
Can you please add a test?

cjwagner:
This change is exactly what I need. Let me know if there is anything I can do to help expedite this PR.

influxdb/client.py
```diff
@@ -219,7 +235,10 @@ def request(self, url, method='GET', params=None, data=None,
         :raises InfluxDBClientError: if the response code is not the
             same as `expected_response_code` and is not a server error code
         """
-        url = "{0}/{1}".format(self._baseurl, url)
+        if self.__uri == '':
+            url = "{0}/{1}".format(self._baseurl, url)
+        else:
+            url = "{0}/{1}/{2}".format(self._baseurl, self.__uri, url)
```
apelisse: Why are you doing this here rather than at `self.__baseurl` initialization?
Also, the condition could be removed if `format` was replaced with `urlparse.urljoin`.

influxdb/client.py
```diff
@@ -62,11 +62,26 @@ class InfluxDBClient(object):
...
+            </LocationMatch>
+
+    :type uri: str
+
     """
 
     def __init__(self,
                  host='localhost',
                  port=8086,
+                 uri='',
```
apelisse: Note that this parameter is not used when `use_udp` is set. But I guess this is fine since that's also true for some other parameters.

influxdb/client.py
```diff
@@ -62,11 +62,26 @@ class InfluxDBClient(object):
     :type udp_port: int
     :param proxies: HTTP(S) proxy to use for Requests, defaults to {}
     :type proxies: dict
+    :param uri: uri to use in URL when using a reverse proxy, defaults to ''
+        example https://myserver.domain.com/influxproxy with an Apache
```
apelisse: I'm not sure the whole proxy configuration is useful here.

GilZm: I was fearing a misunderstanding between the forward proxy and reverse proxy settings... having the configuration here is not mandatory to me.

influxdb/client.py
GilZm: obviously

influxdb/client.py
GilZm: It can be performed at the `self.__baseurl` initialization, your right... when reading the code first time, I focused on the `url` string creation.

sebito91:
A lot of tests are failing with your patch. Could you please update your patch to include:

1. updated patch to pass existing tests
2. updated patch to add new tests for reverse proxy
3. squash your commits into one to ensure clean merge

@aviau UPDATE REQUESTED

sebito91:
Closing stale issue for now. If this is still required please let me know and we can re-open.

Issue #456, opened 2017-05-30
With pandas 0.20.1, DataFrame / write_points call fails with module ‘pandas.tseries’ has no attribute ‘period’ 


lraynaud:
Please, let me know if the below is a known issue. And yes, I use Python 3.6, not officially supported, nonetheless this bug sounds important enough to be reported:    

With pandas 0.19.2, the python-influxDB demo sample given with the Tutorials - pandas works as expected with Python 3.6 (see URL http://influxdb-python.readthedocs.io/en/latest/examples.html). 
I am using Python 3.6.0 |Anaconda custom (x86_64)| (default, Dec 23 2016, 13:19:00)

ipdb> client.write_points(df, 'demo')
True

With pandas 0.20.1, client.write_points fails and the stack trace shows:
ipdb> client.write_points(df, 'demo')
...
_ client.write_points(df, 'demo')_
_ File "/Users/_____/anaconda/lib/python3.6/site-packages/influxdb/dataframe_client.py", line 117, in write_points_
_ numeric_precision=numeric_precision)_
_ File "/Users/_____/anaconda/lib/python3.6/site-packages/influxdb/dataframe_client.py", line 237, in convertdataframe_to_lines_
_ if not (isinstance(dataframe.index, pd.tseries.period.PeriodIndex) or_
AttributeError: module 'pandas.tseries' has no attribute 'period'

Thanks,
Lionel

sebito91:
There are two PR's in now for this! #454 and #457...will update shortly.

PR #454, opened 2017-05-18
[FIX] : compatibility with new version of pandas

patrickhoebeke:
pd.tseries.period.PeriodIndex has been moved to pd.PeriodIndex since at least pandas 0.18.1
pd.tseries.period.DatetimeIndex has been moved to pd.DatetimeIndex since at least pandas 0.18.1

Issue #446, opened 2017-05-08
pd.tseries.period.PeriodIndex has moved with pandas version 20.1


patrickhoebeke:
Hello,

I've just noticed that for some use cases, the library does not work anymore with the new version of pandas v20.1.
In pandas 20.1, stuff related to tseries has been moved/dropped.
So references to "pd.tseries.period.PeriodIndex" should be replaced by "pd.PeriodIndex"

Patrick

sebito91:
Should be fixed on #454 and #457

PR #457, opened 2017-06-02
updating deps for py27, including pandas

sebito91:
Updating deps to latest for influxdb and py27. Will update this thread with py34 and (hopefully) py36.


sebito91:
Dupe of #454 but includes updates to influxdb binary (v1.2.4 instead of v1.1.0).

sebito91:
Fixes #456 and #446.

aviau:
Nice!

PR #458, opened 2017-06-04
adding py35 and py36

sebito91:


sebito91:
Given that we support up to py36 now, should we mark this as Beta instead of Alpha?

aviau:
That is way too many envs builds. Can you please just leave the py35 and py36 envs, not coverage, flake8 and docs?

building docs on one interpreter is fine.

sebito91:
Point taken, fixing now...

aviau:
Looks good to me!

PR #475, opened 2017-06-27
_is_float should return false rather than an error when types are incompatible

BenHewins:
Currently a pandas Datetime column can cause an error to be thrown unnecessarily

### DIFF
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -75,7 +75,7 @@ def quote_literal(value):
 def _is_float(value):
     try:
         float(value)
-    except ValueError:
+    except (TypeError, ValueError):
         return False
 
     return True


sebito91:
This is indeed a good catch, the helper function is somewhat helpful but not enough so! I would prefer we catch the types of errors more gracefully. The code should be made more robust so we capture each of these types as they come up instead of just silently ignore.

Could you please amend this to do something like:

```
In [19]: def _is_float(value):
       :     try:
       :         float(value)
       :     except (TypeError, ValueError):
       :         return False
       :     return True
       :
```

sebito91:
@aviau UPDATED REQUESTED

BenHewins:
@sebito91 I've made the changes you suggested. I'm not quite sure when an exception rather than returning false would be the expected behaviour, but I suppose it's good practice.

sebito91:
Excellent, this looks good! Thanks @BenHewins!

@aviau this is ready to merge!

sebito91:
@BenHewins can you please try to squash into one commit? Thanks!

PR #486, opened 2017-08-03
Improve performance on dataframe to line conversion

maxdolle:
The function influxdb._dataframe_client.DataFrameClient._convert_dataframe_to_lines() is quite slow due when it comes to tags formatting (about 1 second for 10000 lines)

The slowest part was r.dropna() (line 312). I propose the attached solution that combines several steps into a single one to avoid the call to dropna().

### DIFF
--- a/influxdb/_dataframe_client.py
+++ b/influxdb/_dataframe_client.py
@@ -302,17 +302,10 @@ class DataFrameClient(InfluxDBClient):
             tag_df = self._stringify_dataframe(
                 tag_df, numeric_precision, datatype='tag')
 
-            # prepend tag keys
-            tag_df = tag_df.apply(
-                lambda s: s.apply(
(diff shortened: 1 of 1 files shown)


maxdolle:
Here is a small script to the test the performance of the proposed solution.
It has to be tested with influxdb installed (without this patch).

Results on machine:
1.47 seconds without fix
0.08 seconds with fix

[performance_test.txt](https://github.com/influxdata/influxdb-python/files/1197557/performance_test.txt)

sebito91:
@maxdolle could you please squash to one commit? This looks good, thanks for the fix.

@aviau this looks fine to me, should be ready to merge!

maxdolle:
@sebito91 It's squashed!

xginn8:
thanks @maxdolle!

Issue #482, opened 2017-07-27
write seems to fail with Decimal value


ghost:
Hi,

As reported by @chripede in #430 , using Decimal value with write triggers an exception:
e.g if the value is 2695.30
```
influxdb.exceptions.InfluxDBClientError: 400: {"error":"partial write: unable to parse 'totalgain value=Decimal('2695.30')': invalid boolean dropped=0"}
```

See the last message of @chripede in #430 for another example. As he said:

> I think the problem might be that float() will return a float on a Decimal, but repr will actually return "Decimal('29.95')"

Let me know if you need more information.

Best Regards,
Carl

sebito91:
Commented on your #488 PR, why not just return the `repr(float(value))` since you've already passed the check?

sebito91:
Fixed in #813

Issue #431, opened 2017-03-22
precision lost when writing float field values


isilien:
`_escape_value` in `line_protocol.py` loses precision when casting floats to strings. Example:

```
>>> from influxdb import InfluxDBClient
>>> influx = InfluxDBClient(database='test_db')
>>> data_points = [{'fields': {'value': 0.8289445733333332}, 'time': 1488510364159, 'tags': {'foo': 'bar'}, 'measurement': 'test'}]
>>> influx.write_points(data_points, time_precision='ms')
>>> results = influx.query('select * from test')
>>> print list(results)
[[{u'foo': u'bar', u'value': 0.828944573333, u'time': u'2017-03-03T03:06:04.159Z'}]]
```

Pull request for fix: #430

```
Python version: 2.7.6
InfluxDB version: 1.1.1
```

isilien:
Addressed in #430

PR #488, opened 2017-08-11
fix issue with repr and Decimal

chaica:
Given the issue #431 fixing a precision issue while using python 2.x, as demonstrated by the example below:
```
$ python
Python 2.7.13 (default, Jan 19 2017, 14:48:08)
>>> float(repr(0.8289445733333332)) == 0.8289445733333332
True
>>> float(str(0.8289445733333332)) == 0.8289445733333332
False
>>>
```

We have now an issue as described in #430 and #482 with both Python and Python3 and the use of both Decimal Python type and repr, because :
```
>>> from decimal import Decimal
>>> str(Decimal(0.8289445733333332))
'0.828944573333333156739399782964028418064117431640625'
>>> repr(Decimal(0.8289445733333332))
"Decimal('0.828944573333333156739399782964028418064117431640625')"
```
In Python 3, str() and repr() now provides the same result with a float:
```
$ python3
Python 3.5.3+ (default, Jun  7 2017, 23:23:48) 
>>> float(repr(0.8289445733333332)) == 0.8289445733333332
True
>>> float(str(0.8289445733333332)) == 0.8289445733333332
True
>>>
```

But the problems remain while using both Decimal and repr(). The pull request offers a solution which should work both for python2 and python3.

Regards,
Carl

### DIFF
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -90,6 +90,14 @@ def _escape_value(value):
         return str(value) + 'i'
     elif _is_float(value):
         return repr(value)
+        reprvalue = repr(value)
+        decimalstart = "Decimal('"
+        decimalstop = "')"
+        if reprvalue.startswith(decimalstart):
+            return reprvalue.lstrip(decimalstart).rstrip(decimalstop)
+        else:
+            return reprvalue
+
 
     return str(value)
 


sebito91:
Why make this so convoluted, since we're passing this to `_is_float` already (and succeeding), why not just `return repr(float(value))`?

influxdb/line_protocol.py
```diff
@@ -89,7 +89,14 @@ def _escape_value(value):
     elif isinstance(value, integer_types) and not isinstance(value, bool):
         return str(value) + 'i'
     elif _is_float(value):
-        return repr(value)
+        reprvalue = repr(value)
```
sebito91: Why not just make this `return repr(float(value))` since you've already passed the `_is_float` test?

chaica: I guess I went that way because I didn't want to lose precision of the Decimal, but I don't know if it's worth it, your solution is quite shorter and more elegant.

sebito91: If you're referring to your example earlier, even if Decimal resolves to a more granular value than what you sent it may not be what you actually want. If you wanted that value down to the full granularity you should provide that to begin with (IMHO of course).

Do you want to take a stab at changing your PR? If so, can you please add a quick test to check both a non-Decimal and Decimal value?

sebito91: To clarify, since your `Decimal` value is set to 16 digits, you should get the `repr` value of the 16-digit float instead of a (potentially) longer, more granular definition.

```
In [1]: from decimal import Decimal

In [2]: str(Decimal(0.8289445733333332))
Out[2]: '0.828944573333333156739399782964028418064117431640625'

In [3]: repr(Decimal(0.8289445733333332))
Out[3]: "Decimal('0.828944573333333156739399782964028418064117431640625')"

In [4]: repr(float(Decimal(0.8289445733333332)))
Out[4]: '0.8289445733333332'
```

PR #491, opened 2017-08-18
add a test on authtication error

laurentL:


sebito91:
Hmm, this looks to build upon #490, do you want these auth_test changes to that commit?

PR #492, opened 2017-08-19
Data arg in write() can be single string

baftek:
data argument in write() was expected to be list of strings, now it can be single string as well

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -270,7 +270,7 @@ class InfluxDBClient(object):
 
         :param data: the data to be written
         :type data: (if protocol is 'json') dict
-                    (if protocol is 'line') sequence of line protocol strings
+                    (if protocol is 'line') sequence of line protocol strings or single string
         :param params: additional parameters for the request, defaults to None
         :type params: dict
         :param expected_response_code: the expected response code of the write
@@ -284,7 +284,7 @@ class InfluxDBClient(object):
         headers = self._headers
         headers['Content-type'] = 'application/octet-stream'
 
-        if params:
+        if params and 'precision' in params:
             precision = params.get('precision')
         else:
             precision = None
@@ -292,7 +292,10 @@ class InfluxDBClient(object):
         if protocol == 'json':
             data = make_lines(data, precision).encode('utf-8')
         elif protocol == 'line':
-            data = ('\n'.join(data) + '\n').encode('utf-8')
+            if type(data) == list:
+                data = ('\n'.join(data) + '\n').encode('utf-8')
+            elif type(data) == str:
+                data = data.encode('utf-8')
 
         self.request(
             url="write",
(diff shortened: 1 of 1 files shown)


influxdb/client.py
```diff
@@ -284,15 +285,18 @@ def write(self, data, params=None, expected_response_code=204,
         headers = self._headers
         headers['Content-type'] = 'application/octet-stream'
 
-        if params:
+        if params and 'precision' in params:
```
sebito91: We don't need this check here, this will default to None.

```
In [1]: params = {'one': 'two'}

In [2]: if params:
      :     precision = params.get('precision')
      : else:
      :     precision = None
      :

In [3]: precision

In [4]: params
Out[4]: {'one': 'two'}

In [5]: type(precision)
Out[5]: NoneType
```

influxdb/client.py
```diff
@@ -284,15 +285,18 @@ def write(self, data, params=None, expected_response_code=204,
...
             precision = params.get('precision')
         else:
             precision = None
 
         if protocol == 'json':
             data = make_lines(data, precision).encode('utf-8')
         elif protocol == 'line':
-            data = ('\n'.join(data) + '\n').encode('utf-8')
+            if type(data) == list:
```
sebito91: Let's change this slightly as follows:

```
if not isinstance(data, list):
    data = [data]

data = ('\n'.join(data) + '\n').encode('utf-8')
```

baftek: I am not super-experienced Pythonist, however I think that this is not the best solution. Or is it?
Since sequence is requested, user may provide both list or tuple. In this case if tuple is found it will be not casted to list but put into list. Then the following join will fail because instead of str it will find tuple. Cast by list() would be much better, because str will be put into list, and tuple converted into list. Win-win I guess. If any other data structure was provided - exception is justified.

influxdb/client.py
baftek: Well, I thought I got exception from this but apparently missing key with get() gives Null.
You're right.

influxdb/client.py
```diff
@@ -292,6 +293,8 @@ def write(self, data, params=None, expected_response_code=204,
         if protocol == 'json':
             data = make_lines(data, precision).encode('utf-8')
         elif protocol == 'line':
+            if isinstance(data, str):
+                data = list(data)
```
sebito91: This will not do what you expect, instead it'll split out your string into a list as follows:

```
In [4]: tester = "onetwothree"

In [5]: [tester]
Out[5]: ['onetwothree']

In [6]: list(tester)
Out[6]: ['o', 'n', 'e', 't', 'w', 'o', 't', 'h', 'r', 'e', 'e']
```
Please replace with `data = [data]` and that should suffice.

sebito91:
This is great, thank you @baftek!

Issue #479, opened 2017-07-19
DataFrameClient issue - seems does not process correctly DateTimeIndex dates


JohnSamler:
I'm having problems uploading dataframes into InfluxDB, using write_points.

The problem is due to this part of the code:

File _dataframe_client.py

```
        # Make array of timestamp ints
        if isinstance(dataframe.index, pd.PeriodIndex):
            time = ((dataframe.index.to_timestamp().values.astype(int) /
                     precision_factor).astype(int).astype(str))
        else:
            time = ((pd.to_datetime(dataframe.index).values.astype(int) /
                     precision_factor).astype(int).astype(str))
```

That int cast doesn't work in my code. I tested just casting datetime's into ints and this is what i get:

```
    In [24]: numpy.datetime64('2016-01-05T00:00:00.000000000').astype(int)
    Out[24]: -1226113024
    In [25]: numpy.datetime64('2016-01-04T00:00:00.000000000').astype(int)
    Out[25]: 630980608
    In [26]: numpy.version.version
    Out[26]: '1.11.3'
```

I had to modify the code to cast to np.int64 for it to work (not using PeriodIndex):

```
        # Make array of timestamp ints
        if isinstance(dataframe.index, pd.PeriodIndex):
            time = ((dataframe.index.to_timestamp().values.astype(int) /
                     precision_factor).astype(int).astype(str))
        else:
            time = ((pd.to_datetime(dataframe.index).values.astype(np.int64) /
                     precision_factor).astype(np.int64).astype(str))
```

Don't think the cast as coded currently is correct, or a case is missing.

patrickhoebeke:
I confirm the issue.
On my Ubuntu, I do not have the issue, the casting to int generates the correct Epoch required by Influx.
On Windows, the conversion ins rather chaotic if using 'int'.
Casting to np.int64 solves the issue on both OS.

So I propose the following modification:
```python
# Make array of timestamp ints
if isinstance(dataframe.index, pd.PeriodIndex):
    time = ((dataframe.index.to_timestamp().values.astype(int) /
        precision_factor).astype(int).astype(str))
else:
    time = ((pd.to_datetime(dataframe.index).values.astype(np.int64) /
        precision_factor).astype(np.int64).astype(str))
```

As a side note, from a performance point of view, direct casting seems to be more performant. For example with a DataFrame with 1 million rows:
```python
[...]

PR #494, opened 2017-08-23
Fix for  influxdata/influxdb-python#479 : DataFrameClient issue - seems does not process correctly DateTimeIndex dates

patrickhoebeke:
Fix for  influxdata/influxdb-python#479
On windows machine, casting to int does not work properly to convert DateTimeIndex into Unix Epoch (required for Influx). Conversion to np.int64 solves the issue


patrickhoebeke:
sorry, this fix does not solves completely the issue. I'll fix it correctly and post a new merge request. Sorry

PR #495, opened 2017-08-24
Fix for  DataFrameClient issue - seems does not process correctly DateTimeIndex dates (issue #479)

patrickhoebeke:
Fix for  influxdata/influxdb-python#479
Force conversion to 64bits integer (as astype(int) converts to 32bits on some machines and to 64bits on others). 64bits conversion is required.


patrickhoebeke:
Two remaining questions:
1. Is this fix compliant with 32bits machine? If fear that if we want to provide nanoseconds timestamps inputs to Influx, 64bits is a requirement. Then, so should it be part of some check when running the code?
2. I am not sure whether a specific unittest is required

xginn8:
thanks for contributing!

PR #503, opened 2017-09-19
Improve DataFrameClient tags line-protocol conversion performance

tzonghao:
In `_convert_dataframe_to_lines`, if only `global_tags` is specified but
not `tag_columns`, take a faster route to process the tags. Previously,
in such a case, global tags are duplicated as tag columns and processed
as if they were tag columns. Such processing is wasteful and results in
a slowdown that becomes noticeable when batch loading many thousands of
data points with a handful of global tags.


aviau:
Nice!

PR #507, opened 2017-09-28
Fix issues #422 and #479

patrickhoebeke:
[FIX] :  Fix issue #422 
with this implementation, now DataFrameClient allow to push DataFrame with null values (NaNs or None)
- **Issue**: DataframeClient fails to push DataFrame with NaNs
- **Root cause** : null are not allowed on Influx (null is not a valid type)
- **Possible work around**: push column per column (with dropnan() before each push)
- **Solution** : corresponding points are dropped from the list of points pushed to influx. Also takes into account full lines with nulls and DataFrame full of nulls)

[FIX] : Fix issue #479 
conversion of DatetimeIndex to epoch was not correct on some machines. 
- **Issue** : on some Windows machines, DatetimeIndex not correctly converted to epoch
- **Root cause**: to convert to Epoch, timestamp is first converted to int. On some machines, default conversion to int leads to a conversion to int32 (which lead to incorrect conversion to epoch). Conversion to int should be to int64.
- **Solution**: conversion forced to int64 in the code


influxdb/__init__.py
```diff
@@ -18,4 +18,4 @@
 ]
 
 
-__version__ = '4.1.1'
+__version__ = '4.1.1b'
```
aviau: please don't bump the version in pull requests.

tux-00:
In commit f40f77065ca49334db3e3f587673d3844d925f37 you use dropna() but this method seems to delete all the row (https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.dropna.html).

Maybe I'm wrong but I use this code to drop NaN values : `df.fillna('')` instead of `dropna()`.

sebito91:
Closing stale request, addressed in later release.

Issue #416, opened 2017-02-14
Wrap exception in InfluxDBClientError


Don42:
When trying to write  to influx and the database is not available a requests ConnectionError is raised.
This happens after three retries at this location [0]. This exception should be wrapped or replaced with an InfluxDBClientError to simplify error handling for the user of the library.

[0] https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L246

PierreF:
I think InfluxDBServerError is more appropriate.

Don42:
Well, it might be an error in the server (broken/down) or the client (configuration issue) or somewhere in between (network). Maybe add a superclass for all exceptions the module might raise and use that when you can't really define if it's a client or server issue.

renderit:
+1

Don42:
This isn't really fixed by #508. My issue is that a user of your library should not need to know what library you are using to write to the database to handle all errors. Currently I have to import requests myself to handle connection exceptions.

sogrom:
> This isn't really fixed by #508. My issue is that a user of your library should not need to know what library you are using to write to the database to handle all errors. Currently I have to import requests myself to handle connection exceptions.

Totally agree, please wrap raise requests.exceptions.ConnectionError in InfluxDBServerError
User should not have to study your code to know that you are raisng some imported exceptions. 

In doc there are only two exceptions, that can be raised in your lib.
https://influxdb-python.readthedocs.io/en/latest/exceptions.html

Issue #378, opened 2016-10-19
DataFrameClient retry on batches


dirkdevriendt:
When writing medium to large dataframes (multiple 50K datapoints and up) we sometimes run into server errors (HTTP 500, timeout etc). I understand this could be considered outside of the scope for the client library, but if the error is caught outside of the batch loop we can't know what data was uploaded and have to restart the upload for the entire dataframe.

Ideally, when using the the batch_size option on the DataFrameClient.write_points function it would be aware of these errors and retry a batch if needed. There are a few retry decorators out there that could be of use (e.g. https://github.com/d0ugal/retrace).

Any thoughts? Would this be a valuable addition? Would you mind taking a dependency on a retry decorator?

geethanjalieswaran:
+1

rterbush:
+1, just running into this issue

500 error is leaving influxdb server spinning off and taking the server down with it as a result.

rafaelcapucho:
+1

xginn8:
I added retry logic on ConnectionError a few months back, I'll implement something similar for HTTPErrors with exponential backoff (see https://github.com/influxdata/influxdb-python/pull/435).

PR #508, opened 2017-09-29
adding retry logic for all requests.exceptions

xginn8:
adding an exponential backoff with jitter on POST calls
reducing the number of retries in some of the tests to comply with the
backoff feature.
pass through all requests.exceptions after the permitted number of retries.

fixes https://github.com/influxdata/influxdb-python/issues/378, maybe could fix https://github.com/influxdata/influxdb-python/issues/416 too but I'm not sure about how best to pass back requests.exceptions.


xginn8:
new build failure looks to be the same travis issue as https://github.com/aayush26/pirant/issues/19

PR #511, opened 2017-09-30
Fix failing tags match in get_points() on a ResultSet

### LINKED ISSUE #506: Filtering by tags results in empty list
I'm observing an unexpected issue while reading data from InfluxDB

```
client = InfluxDBClient('localhost', 8086, 'root', 'root', 'influx_demo')

result = client.query("SELECT * FROM energy WHERE time >= '2015-01-01T00:00:00Z' AND time <= '2015-01-01T00:30:00Z'")

series = list(result.get_points(tags={'source': 'SBERA0'}))
```
series variable is an empty list and it should not; the same query using CLI returns the following points where [account location  source] are "Tags" and value is "Field"
```
> SELECT * FROM energy WHERE time >= '2015-01-01T00:00:00Z' AND time <= '2015-01-01T00:30:00Z'
name: energy
time                account location source value
----                ------- -------- ------ -----
1420070400000000000 BER     A        SBERA0 1
1420070400000000000 PAR     A        SPARA0 3
1420070400000000000 LOS     A        SLOSA0 6
1420070400000000000 LON     A        SLONA0 4
1420070400000000000 MIL     A        SMILA0 4
1420070700000000000 LON     A        SLONA0 6
1420070700000000000 MIL     A        SMILA0 8
[...]
```

I can make it work by using GROUP BY clause
```
result = client.query("SELECT * FROM energy WHERE time >= '2015-01-01T00:00:00Z' AND time <= '2015-01-01T00:30:00Z' GROUP BY source")

series = list(result.get_points(tags={'source': 'SBERA0'}))
```
But at this point I can only use the tag you have used in GROUP BY tags={'source': <value>} 
and using  multiple tags  tags={'source': <value>, 'location': <value>} is not returning appropriate values.

xginn8:
fixes #506.

Most of this diff is renaming or reformatting, the only real logical change is at line 108 where I flipped the if & for statements, and passed the tags into _tag_matches rather than passing an empty dict.


xginn8:
fixing CI now, i omitted server tests from my testing.

aviau:
Can you explain this to me a little?

Does that mean that tags no longer exist in influxdb?

aviau:
Looking at the API on the influxdb website, this looks good to me.

PR #513, opened 2017-10-10
Fixes typo in Content-Type header

milancermak:
This PR fixes the name of the `Content-Type` header.

Even though requests uses a case-insensitive dict for storing HTTP headers, I think it's better form to use [proper naming](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html).


aviau:
Thanks!

PR #518, opened 2017-10-17
Added support for not including all fields when using the series helper

### LINKED ISSUE #302: Why all fields must be present in SeriesHelper?
```
class SeriesHelper(object):
    def __init__(self, **kw):                                                 
        """
        Constructor call creates a new data point. All fields must be present.
        """
```

Why _SeriesHelper_ has such constraint? Fields in InfluxDB are not indexed. Data points can be spatial in fields to ignore some missing values.

spott:
Fixes #302.


spott:
I don't think the PyPy error is unique to this pull request...

xginn8:
Thanks for contributing! Would you mind adding a simple test, such that the NameError you raise is caught?

spott:
Is there anything else that is needed?

Issue #349, opened 2016-07-04
Connection pool is full, discarding connection


newtoncorp:
When using influxdb-python in code that uses concurrent.futures.ThreadPoolExecutor heavily, the connections are not reaped and at some point, the data is discarded because there are too many different connections.

wolong12:
I had the same problem when writing data to influxdb. I wrote 20 points at almost the same time, within different requests. Default connection pool size by 10 was not enough in this case.

TeraHz:
Same issue when writing a lot of points from Home Assistant. It exhausts the connection pool very quickly (immediately)

alyan2008:
Hi

Seems I have the same problem. Did you fix it?

lesingerouge:
This seems like a requests limitation when handling the urllib3 connection pool manager.
It's pretty much hardcoded here: https://github.com/kennethreitz/requests/blob/master/requests/adapters.py

sharkman-dev:
I had the same problem. Maybe some additional info can help to reproduce?

mihalski:
I've also run into this issue.
So does this effectively make InfluxDB totally useless for Home Assistant? Or is there another way to connect to it from Home Assistant?

TeraHz:
My Issue has gone away since HA changed the schema. I think it happened around release 36?

mihalski:
I'm getting a lot of:
```
Traceback (most recent call last):
  File "/usr/lib/python3.4/concurrent/futures/thread.py", line 54, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.4/site-packages/homeassistant/components/influxdb.py", line 155, in influx_event_listener
    influx.write_points(json_body)
  File "/home/homeassistant/.homeassistant/deps/influxdb/client.py", line 402, in write_points
    tags=tags)
  File "/home/homeassistant/.homeassistant/deps/influxdb/client.py", line 447, in _write_points
    expected_response_code=204
  File "/home/homeassistant/.homeassistant/deps/influxdb/client.py", line 289, in write
    headers=headers
  File "/home/homeassistant/.homeassistant/deps/influxdb/client.py", line 245, in request
    timeout=self._timeout
  File "/srv/homeassistant/homeassistant_venv/lib/python3.4/site-packages/requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
[...]

PR #534, opened 2017-11-06
Add pool size parameter to client constructor

vaniakov:
This pull request relates to the issue [Connection pool is full, discarding connection](https://github.com/influxdata/influxdb-python/issues/349). To reproduce it you need to create more than 10 (default size of the urllib3 connection pool) simultaneous connection to the database from the same client object.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -33,11 +33,9 @@ else:
 
 class InfluxDBClient(object):
     """InfluxDBClient primary client object to connect InfluxDB.
-
     The :class:`~.InfluxDBClient` object holds information necessary to
     connect to InfluxDB. Requests can be made to InfluxDB directly through
     the client.
-
     :param host: hostname to connect to InfluxDB, defaults to 'localhost'
     :type host: str
     :param port: port to connect to InfluxDB, defaults to 8086
@@ -46,6 +44,8 @@ class InfluxDBClient(object):
     :type username: str
     :param password: password of the user, defaults to 'root'
     :type password: str
+    :param pool_size: urllib3 connection pool size, defaults to 10.
+    :type pool_size: int
     :param database: database name to connect to, defaults to None
     :type database: str
     :param ssl: use https instead of http to connect to InfluxDB, defaults to
@@ -73,6 +73,7 @@ class InfluxDBClient(object):
                  port=8086,
                  username='root',
                  password='root',
+                 pool_size=10,
                  database=None,
                  ssl=False,
                  verify_ssl=False,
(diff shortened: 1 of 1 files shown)


influxdb/client.py
```diff
@@ -64,6 +66,7 @@ def __init__(self,
                  port=8086,
                  username='root',
                  password='root',
+                 pool_size=10,
```
xginn8: I think it makes sense to cast this value to an int, like we do for the `port` parameter in line 81.

vaniakov: @xginn8 make sense.

vaniakov:
@xginn8 nice catch!

xginn8:
thanks for contributing!

hppylbb:
Hi all, sorry but I find this issue is still exist in the master branch, the parameter pool_size is invalid even we added it to \_\_init\_\_ method. I think that https://github.com/influxdata/influxdb-python/blob/35732cd7dfe5a585564999c9f881bd88e7c1531d/influxdb/client.py#L106 should be modified to "`self._session.mount(self._scheme+'://', adapter)`",
because after executing this line, the "session.adapters" will have three members: "`OrderedDict([('https://', <requests.adapters.HTTPAdapter object at 0x7f0c9c123890>), ('http://', <requests.adapters.HTTPAdapter object at 0x7f0c9404e190>), ('http', <requests.adapters.HTTPAdapter object at 0x7f0c9e874290>)]))`", the https:// and http:// adapters are added here https://github.com/requests/requests/blob/3dc84cde5b785a85cb20eb2f0d4530fd568a6af7/requests/sessions.py#L391-L393 while initializing, and when you execute https://github.com/requests/requests/blob/3dc84cde5b785a85cb20eb2f0d4530fd568a6af7/requests/sessions.py#L691, you want to get the adapter 'http' which is mounted at https://github.com/influxdata/influxdb-python/blob/35732cd7dfe5a585564999c9f881bd88e7c1531d/influxdb/client.py#L106 (the third one in OrderedDict),  but here you got a incorrect adapter which is http:// (the second one in OrderedDict), this adapter doesn't contain the 'pool_size' parameter we added to \_\_init\_\_ method while initializing InfluxDBClient.

So while we write_points to influxdb, we still use the default pool_size=10, then we run into issue https://github.com/influxdata/influxdb-python/issues/349.

If we add '://' to `self._session.mount(self._scheme+'://', adapter)`, then the OrderedDict will only contains two adapters: `OrderedDict([('https://', <requests.adapters.HTTPAdapter object at 0x7f0c9c123890>), ('http://', <requests.adapters.HTTPAdapter object at 0x7f0c9404e190>)])`, the http:// adapter will be overrided by our session mount method, then we can get the correct http adapter with the pool_size we added.

vaniakov:
@hppylbb you are correct. Working on the new PR.

PR #536, opened 2017-11-09
Add some small improvements

swails:
- six is already listed as a requirement, so use six.moves instead of
defining builtins based on checking sys.version_info
- Fix formatting of several exceptions.

### DIFF
--- a/influxdb/_dataframe_client.py
+++ b/influxdb/_dataframe_client.py
@@ -222,8 +222,8 @@ class DataFrameClient(InfluxDBClient):
                             .format(type(dataframe)))
         if not (isinstance(dataframe.index, pd.PeriodIndex) or
                 isinstance(dataframe.index, pd.DatetimeIndex)):
-            raise TypeError('Must be DataFrame with DatetimeIndex or \
-                            PeriodIndex.')
+            raise TypeError('Must be DataFrame with DatetimeIndex or '
+                            'PeriodIndex.')
 
         # Make sure tags and tag columns are correctly typed
         tag_columns = tag_columns if tag_columns is not None else []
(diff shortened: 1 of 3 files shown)


influxdb/_dataframe_client.py
```diff
@@ -222,8 +222,8 @@ def _convert_dataframe_to_json(dataframe,
                             .format(type(dataframe)))
         if not (isinstance(dataframe.index, pd.PeriodIndex) or
                 isinstance(dataframe.index, pd.DatetimeIndex)):
-            raise TypeError('Must be DataFrame with DatetimeIndex or \
-                            PeriodIndex.')
+            raise TypeError('Must be DataFrame with DatetimeIndex or '
+                            'PeriodIndex.')
```
swails: If this was ever raised, the formatting would end up looking weird:

```python
>>> raise TypeError('this is a long error message \
...                 that gets split over lines, but look \
...                 at all those spaces!')
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
TypeError: this is a long error message                 that gets split over lines, but look                 at all those spaces!
```

influxdb/client.py
```diff
@@ -254,25 +243,21 @@ def request(self, url, method='GET', params=None, data=None,
...
+                    requests.exceptions.Timeout):
                 _try += 1
                 if self._retries != 0:
                     retry = _try < self._retries
                 if method == "POST":
                     time.sleep((2 ** _try) * random.random() / 100.0)
-        if _error:
-            raise(_error)
+                if not retry:
+                    raise
```
swails: linters like this approach a little better, and it also gives more digestible tracebacks IMO:

```python
>>> e = None
>>> try:
...     raise ValueError('this is a raised value error')
... except ValueError as _e:
...     e = _e
... 
>>> raise(e)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in <module>
ValueError: this is a raised value error
```

```python
>>> try:
...     raise ValueError('this is a raised value error')
... except ValueError:
...     raise
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
ValueError: this is a raised value error
```

This seems to be the original intent.

PR #538, opened 2017-11-26
Support streamed chunking (fixing #531 and others)

### LINKED ISSUE #531: Out of memory error when querying large data set with chunked=True
Making large queries for very large result sets results in out of memory errors.  This can be traced back to the function _read_chunked_response in client.py which reads the entire response into RAM before returning a ResultSet.

psy0rz:
query() will now return a generator of ResultSets instead of a list when setting chunked=True. This might be backwards incompatible, but seems to be the only reason one might even use chunked mode?

Another reason to use chunked mode could be that the server enforces it, but then again: If results are that big, wouldn't you want a generator as well?

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -387,7 +387,7 @@ class InfluxDBClient(object):
         :type chunk_size: int
 
         :param stream: Will stream the data and return a generator that
-            generates one ResultSet per chunk containing.
+            generates one ResultSet per chunk.
             This allows for huge datasets with virtually no limit.
 
         :type stream: bool
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -201,7 +201,7 @@ class InfluxDBClient(object):
         self._password = password
 
     def request(self, url, method='GET', params=None, data=None,
-                expected_response_code=200, headers=None, chunked=False):
+                expected_response_code=200, headers=None, stream=True):
         """Make a HTTP request to the InfluxDB API.
 
         :param url: the path of the HTTP request, e.g. write, query, etc.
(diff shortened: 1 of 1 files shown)


psy0rz:
i'll try to fix the test as well

psy0rz:
I think it would be cleaner to add a "stream" parameter to get a streaming response. This way we wont breakn compatibility. Perhaps there are also some usecases where someone wants to autonagically combine a chunked result in memory, like it is currently working.

psy0rz:
All set and done now. there a new stream parameter to support "streamed chunking". this is usefull for very large result sets. (trying to traverse and anaylse over 20 million+ results here) Also updated/added documentation in the sourcecode.

related project: https://github.com/psy0rz/meowton

xginn8:
@psy0rz thank you for the patch! would you mind adding a test to this PR?

psy0rz:
Ok will do that soonish..

joaopcbertoldo:
Hi, I believe that someone forgot to actually resolve the conflicts.

@xginn8 do you mind finishing this merge PR?

psy0rz:
also i dont have time to add the requested regressions tests.

aviau:
If the author does not have the time to finish the PR then someone has to pick it up. I'll close this one and merge the other, if it is complete and includes tests.

psy0rz:
"Potentially break the api" ;) Dont worry you can be harsh, I would apply the same policy in your case.

aviau:
I had picked that up from your initial comment saying it could be backwards incompatiblte but this is not t he case due to the steam parameter, is it?

psy0rz:
yes, but still i could have made a mistake :)

pedrofaria09:
Any solution for this? returning a query with millions of rows will affect memory badly.

xiandong79:
returning a query with millions of rows is still impossible on 2019-08-01. @pedrofaria09 @aviau

Issue #497, opened 2017-09-01
Can not work, issue show KeyError: '[0] not in index', need your great help .thanks.


leonyan1314:
Create pandas DataFrame
Create database: demo
Write DataFrame
Traceback (most recent call last):
  File "/Users/LeonYan/MyPython/QI/DF2influx.py", line 50, in <module>
    main(host=args.host, port=args.port)
  File "/Users/LeonYan/MyPython/QI/DF2influx.py", line 25, in main
    client.write_points(df, 'demo', protocol=protocol)
  File "/Users/LeonYan/Library/Python/2.7/lib/python/site-packages/influxdb/_dataframe_client.py", line 125, in write_points
    field_columns=field_columns)
  File "/Users/LeonYan/Library/Python/2.7/lib/python/site-packages/influxdb/_dataframe_client.py", line 220, in _convert_dataframe_to_json
    dataframe[field_columns].to_dict('record'))
  File "/Users/LeonYan/Library/Python/2.7/lib/python/site-packages/pandas/core/frame.py", line 1958, in __getitem__
    return self._getitem_array(key)
  File "/Users/LeonYan/Library/Python/2.7/lib/python/site-packages/pandas/core/frame.py", line 2002, in _getitem_array
    indexer = self.loc._convert_to_indexer(key, axis=1)
  File "/Users/LeonYan/Library/Python/2.7/lib/python/site-packages/pandas/core/indexing.py", line 1231, in _convert_to_indexer
    raise KeyError('%s not in index' % objarr[mask])
KeyError: '[0] not in index'

leonyan1314:
MAC OS 10.12.1,python 2.7

eric-guanwy:
I don't know whether you have got it resolved by now, and I just encountered the same keyerror issue today, for my situation it fixed by specifying the  attribute "columns" explicitly with a string type '0' when create a DataFrame.  As the code below:
  `  df = pd.DataFrame(data=list(range(30)),index=pd.date_range(start='2014-11-16',periods=30, freq='H'),columns=['0'])
`

kiranmantri:
In fact, you can assign any string-based name to your columns.
If `columns=['something'] is suppressed, pandas assigns numerical id them.

*work: *
 `df = pd.DataFrame(data=list(range(30)),index=pd.date_range(start='2014-11-16',periods=30, freq='H'),columns=['0'])`

*doesn't work:*
[...]

PR #547, opened 2017-12-27
Fix pandas example

techaddicted:
Current example code raises KeyError: '[0] not in index', which can
be fixed by adding columns=['0'] to the DataFrame creation.
Fixes issue #497


bonaime:
could you merge this pull-request ?

PR #555, opened 2018-01-12
Make the pandas client easy to use, and fix the precision invalids

SatoshiNewton:
Here, I just make the pandas dataframe client convert the key to string automatically.
Further more, I found the example have a problem about time\_precision.
The time\_precision of dataframe client required the lower word, but pandas automatically make the freq in upper word.
I am not pretty sure did that's relative with the update of pandas, but I fixed it with consider about upper word and lower word at the same time.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -492,7 +492,7 @@ class InfluxDBClient(object):
                       retention_policy,
                       tags,
                       protocol='json'):
-        if time_precision not in ['n', 'u', 'ms', 's', 'm', 'h', None]:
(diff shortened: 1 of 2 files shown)


influxdb/client.py
```diff
@@ -492,7 +492,7 @@ def _write_points(self,
                       retention_policy,
                       tags,
                       protocol='json'):
-        if time_precision not in ['n', 'u', 'ms', 's', 'm', 'h', None]:
+        if time_precision not in ['n', 'u', 'ms', 's', 'm', 'h', 'N', 'U', 'L', 'S', 'M', 'H', None]:
```
sebito91, line 495: I'm not sure what these extra time precision components are meant to be?

sebito91:
Closing for now, please let me know if this is an important fix and we can re-open!

PR #556, opened 2018-01-15
Allow connecting to influxdb running on a path on the server

gladhorn:
Make it possible to connect to the databases on a path on servers.
https://someserver.com/myinfluxdb instead of the root of the server.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -59,6 +59,8 @@ class InfluxDBClient(object):
     :type udp_port: int
     :param proxies: HTTP(S) proxy to use for Requests, defaults to {}
     :type proxies: dict
+    :param path: path of InfluxDB on the server to connect, defaults to ''
+    :type path: str
     """
 
     def __init__(self,
@@ -75,10 +77,12 @@ class InfluxDBClient(object):
                  udp_port=4444,
                  proxies=None,
                  pool_size=10,
+                 path='',
                  ):
         """Construct a new InfluxDBClient object."""
         self.__host = host
         self.__port = int(port)
+        self.__path = path if not path or path[0] == '/' else '/' + path
         self._username = username
         self._password = password
         self._database = database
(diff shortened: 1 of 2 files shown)


gladhorn:
As far as I can tell the failures in the tests are unrelated in this change and also visible in other merge requests.

gladhorn:
Is anyone looking at these requests? And is there some way to re-run travis?

xginn8:
sorry about the delay -- looking into CI failures.

gladhorn:
Thanks :)

influxdb/client.py
```diff
@@ -75,10 +77,12 @@ def __init__(self,
                  udp_port=4444,
                  proxies=None,
                  pool_size=10,
+                 path='',
                  ):
         """Construct a new InfluxDBClient object."""
         self.__host = host
         self.__port = int(port)
+        self.__path = path if not path or path[0] == '/' else '/' + path
```
xginn8: Since you're slicing the string here, perhaps it would make sense to first cast the input to a string (similar to the above port cast)?

gladhorn: So you'e prefer something like:
```
_path = str(path)
self.__path = _path if not _path or _path[0] == '/' else '/' + _path
```
Any other suggestion how to make this look better?

aviau: It dosen't have to be a oneliner

gladhorn: Sure, my suggestion was no longer a one-liner either. What should it be like?

micw:
Any chance to get this reviewed soon? I stumbled across this issue in HomeAssistant which uses the library and I'd like to create a patch for it - but that needs this feature released ;)

Kind regards,
Michael.

gladhorn:
path=None is tested in the test already, it's the default case. The newly added tests test the other cases.

sthenault:
would be interested in this PR too

gladhorn:
Thanks for taking it in :) less ugly workarounds on my side.

Issue #554, opened 2018-01-11
UDP Precision still locked at seconds


mdhausman:
Debugging some issues getting InfluxDB metrics from our python services, I found that even [in the latest code](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L500) UDP writes are locked into seconds precision when this hasn't been a necessity for influxdb since [early 2016](https://github.com/influxdata/influxdb/pull/5565).

Isn't it time for this to be updated? I'd be happy to jot up a quick pull request for it.

aviau:
Feel free to open a PR

xginn8:
Fixed by #557

PR #557, opened 2018-01-15
Remove UDP Precision Restrictions

mdhausman:
- address issue #554
- UDP writes can now convert timestamps to the desired precision
    - add time_precision argument to `InfluxDBClient.send_packet`, defaults to None
- add tests for udp precision
    - remove old udp precision failure tests

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -497,11 +497,6 @@ class InfluxDBClient(object):
                 "Invalid time precision is given. "
                 "(use 'n', 'u', 'ms', 's', 'm' or 'h')")
 
-        if self._use_udp and time_precision and time_precision != 's':
-            raise ValueError(
-                "InfluxDB only supports seconds precision for udp writes"
-            )
-
         if protocol == 'json':
             data = {
                 'points': points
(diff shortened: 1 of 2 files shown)


mdhausman:
So... I tracked down the bug that's causing tests to break, but I'm perplexed because it doesn't look to be related to my change. It _should_ be broken on `master` as well. 

Unless I'm misunderstanding, it's breaking [here](https://github.com/influxdata/influxdb-python/blob/master/influxdb/tests/dataframe_client_test.py#L402) when it tries to send the dataframe in json format. In json, the points get processed through [`_convert_dataframe_to_json`](https://github.com/influxdata/influxdb-python/blob/master/influxdb/_dataframe_client.py#L215) which, unlike `_convert_dataframe_to_line`, lacks a `numeric_precision` arg. Fields are pulled directly from columns with no logic involving numeric precision (e.g. [`_stringify_dataframe`](https://github.com/influxdata/influxdb-python/blob/master/influxdb/_dataframe_client.py#L384)) at all. I feel like I must be misunderstanding, because I don't see how this test could have succeeded in the first place as written.

mdhausman:
Okay, I can see the same test failing in #556 and #555. I have to assume something it just out of whack here unrelated to my change.

mdhausman:
Tagging @aviau for awareness

xginn8:
Thanks for your contribution! Would you mind fixing the flake8 lint failures? Looks good otherwise.

mdhausman:
Did the thing!

PR #562, opened 2018-02-08
Add keyword to set the certificate files for ssl client authentication

goriccardo:
This simple patch allow the client to pass a certificate to the server for ssl client authentication.

We use SSL client authentication extensively for many services and it would be great if the python influx client would support it out of the box.


goriccardo:
Implemented in #702

PR #565, opened 2018-03-03
added endpoint parameter

fradaloisio:
This pull request will add the possibility to have influxdb behind a reverse proxy and using an endpoint to route the requests to it.
A possible use case could be https://myhost.com/influxdb .


fradaloisio:
I've missed the #556 . Sorry about that.

PR #571, opened 2018-03-26
Fix wrong session mount

vaniakov:
Fix for issue addressed in #534.


vaniakov:
@xginn8  could you please take a look at this?

xginn8:
Thanks for contributing!

PR #581, opened 2018-04-27
Remove comment as issues have been resolved

dmuiruri:
It seems the instruction to avoid using the line protocol was created when there were issues that made the line protocol to fail. The issues have been since fixed and the line protocol works fine.


aviau:
Thank you for your contribution!

Issue #583, opened 2018-05-02
Spaces in a dataframe column names breaks the line protocol


dmuiruri:
An issue has been observed where a received dataframe contains un-escaped spaces in the column names (fields or tag keys), this results in a broken line protocol.

To reproduce:

`data = pd.DataFrame(np.random.random((10, 4)), columns=['a', 'b', 'First Entry', 'Second Entry'], index=pd.date_range(start='01/01/2000', periods=10))`
`client = DataFrameClient(database_name)`
`client.write_points(data, 'sample', protocol=line)`

Other solved issues related to this one: #520 #542

xginn8:
Fixed by #584

PR #584, opened 2018-05-03
Parse column names in a dataframe to avoid breaking the line protocol

dmuiruri:
If a dataframe contains spaces in the column names i.e tags and field keys, the spaces cause an error when the line protocol is applied, the spaces are however well handled when the protocol applied is json.

This is a fix to issue #583 .

### DIFF
--- a/examples/tutorial_pandas.py
+++ b/examples/tutorial_pandas.py
@@ -12,7 +12,6 @@ def main(host='localhost', port=8086):
     user = 'root'
     password = 'root'
     dbname = 'demo'
-    # Temporarily avoid line protocol time conversion issues #412, #426, #431.
     protocol = 'json'
 
     client = DataFrameClient(host, port, user, password, dbname)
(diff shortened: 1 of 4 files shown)


aviau:
Please add a test  for whatever you are trying to fix

dmuiruri:
Is there a way to test against a previous release i.e one without the fix, just to see if the test case would catch the error

aviau:
Yes, you can test against a previous release by undoing your code and running the test.

dmuiruri:
Something seems to be going wrong on the server side : 

> ERROR:   flake8: commands failed

dmuiruri:
Build still fails due to what seems like flake8 environment issues on the server side.

dmuiruri:
Any ideas how to resolve the flake8 errors?

xginn8:
@dmuiruri i'll take a look at the failing flake8 run.

xginn8:
Thanks for contributing!

PR #585, opened 2018-05-04
Fixed/Implemented Chunked Read.

gopalkpd:
The document mentions that InfluxDBClient.Query method returns one ResultSet per data chunk
if chunking is enabled by setting the chunked parameter to True. But only one ResultSet containing all
data rows was being returned. This in effect is like having no chunking and retrieving all data in one shot.
Also, for chunked data read to work, streaming should be enabled on the underlying requests.Session.request method
by setting the stream parameter to True. This was not being done.

Made the following changes to make the chunked read work as documented.

*	Updated the method _read_chunked_response to make it a generator that yields
	one ResultSet per data chunk.

*	Enabling streaming on requests.Session.request by passing the parameter “stream=True”
	whenever chunking is enabled on InfluxDBClient.Query.

*	Updated the entire call chain from InfluxDBClient.Query to requests.Session.request method
	to support streaming by adding the parameter “stream”.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -200,7 +200,7 @@ class InfluxDBClient(object):
         self._username = username
         self._password = password
 
-    def request(self, url, method='GET', params=None, data=None,
+    def request(self, url, method='GET', params=None, data=None, stream=False,
                 expected_response_code=200, headers=None):
         """Make a HTTP request to the InfluxDB API.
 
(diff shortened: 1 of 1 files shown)


psy0rz:
Isnt this trying to fix the same thing as https://github.com/influxdata/influxdb-python/pull/538 ?

sebito91:
Duplicate of #538.

PR #588, opened 2018-05-23
#1 - Since pandas 0.23, index.to_datetime() should be changed to to_p…

museghost:
According to the changes in pandas 0.23, index.to_datetime() should be changed to to_pydatetime() in the _dateframeclient.py

https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DatetimeIndex.html


xginn8:
Fixed by #593

Issue #591, opened 2018-06-01
Performance degradation with line protocol on master vs. v5.0.0


shushen:
On master, commit bf232a7a introduced the much welcomed support of dropping NaN/null fields. But it has done so by assembling line by line with function `format_line()`.

This has degraded the performance significantly. See profiling result attached diagram of how `format_line()` is taking the majority of consumed time. 
![performance degradation](https://user-images.githubusercontent.com/870631/40859221-e9693132-6595-11e8-8bcb-58a75c418269.png)
 

Writing a 86400-row dataframe that took only about <10 seconds with v5.0.0 now takes about 90 seconds in my test.

PR #592, opened 2018-06-01
Fix performance degradation with line protocol

shushen:
Assembling line by line in the commit bf232a7aef to remove NaN has
significant performance impact.

This change fixes the issue by keeping the NaN fields before stringify
the dataframe, replacing the fields with empty string, and reverting
back to use pd.DataFrame.sum() function to yield the lines.

Fixes: #591

### DIFF
--- a/influxdb/_dataframe_client.py
+++ b/influxdb/_dataframe_client.py
@@ -365,16 +365,18 @@ class DataFrameClient(InfluxDBClient):
 
         # Make an array of formatted field keys and values
         field_df = dataframe[field_columns]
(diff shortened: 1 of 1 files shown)


shushen:
Unfortunately v5.1.0 has just been released with the performance regression vs v5.0 as addressed here. 

@xginn8 Could you please kindly take a look at this PR?

aviau:
I'll wait for his input since he has been leading the data frame client but I will take a look if he does not respond.

Thanks for contributing and sorry for the wait :)

PR #593, opened 2018-06-07
Update _dataframe_client.py

dragoshenron:
Permanent error "NameError: name 'to_datetime' is not defined".
Reason: to_datetime is a function defined in pandas not a method on a DataFrame. (see: https://stackoverflow.com/questions/48387878/attributeerror-dataframe-object-has-no-attribute-to-datetime)
Tested with Python 3.5.3 and Pandas 0.23.0


xginn8:
Thank you for contributing!

jflory7:
Awesome – I just hit this issue too and saw it was just fixed. This would be awesome to get into a new bugfix release on PyPI.

PR #596, opened 2018-06-26
bump pandas and numpy

aviau:
@xginn8 test_write_points_from_dataframe_with_numeric_precision fails with the new pandas.

Any idea why?

Feel free to commit on this branch btw!

\\cc @sebito91  if you are interested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ addons:
     - wget
 
 matrix:
-  allow_failures:
-    - python: 3.4
-      env: TOX_ENV=docs
   include:
     - python: 2.7
       env: TOX_ENV=py27
@@ -16,8 +13,6 @@ matrix:
       env: TOX_ENV=pep257
     - python: pypy-5.3.1
       env: TOX_ENV=pypy
-    - python: 3.4
-      env: TOX_ENV=py34
     - python: 3.5
       env: TOX_ENV=py35
     - python: 3.6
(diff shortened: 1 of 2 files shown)


xginn8:
Thanks for the heads up, I'll take a look.

xginn8:
Looks like there are a few issues here. First, for Python3.4 there is no official pandas release (http://pandas.pydata.org/pandas-docs/version/0.23/install.html#python-version-support). In my testing, the regression only appears beginning with numpy 0.14.0 (even with pandas 0.23.1 the tests pass). I need to dig a bit more as to why the regression is happening.

@aviau are you amenable to dropping Python3.4 support?

aviau:
Yep, I am good with dropping 3.4

xginn8:
There's a note in the 1.14.0 numpy release notes which is what I believe we've run into (from https://docs.scipy.org/doc/numpy/release.html#compatibility-notes):

`
The str and repr of ndarrays and numpy scalars have been changed in a variety of ways. These changes are likely to break downstream user’s doctests.
`

There's a further note suggesting a possible workaround:
`
Some of these changes are described in more detail below. If you need to retain the previous behavior for doctests or other reasons, you may want to do something like:
`
`
# FIXME: Set numpy array str/repr to legacy behaviour on numpy > 1.13
try:
    np.set_printoptions(legacy='1.13')
except TypeError:
    pass
`
I'll work on an actual fix this week, but if we want to unblock this immediately I've tested that workaround successfully.

PR #598, opened 2018-07-01
Add a changelog with stubs for older releases

### LINKED ISSUE #499: Implement CHANGELOG
Would be nice to cut releases on a regular cadence, adding this to track progress on a release cycle + process.

xginn8:
For now, just annotate the past two releases. Fixes #499


aviau:
Good idea!

Sometimes I forget to mention stuff in the tags, it will help to document our progress in the changelog as we go.

PR #602, opened 2018-07-03
Use default DB in InfluxDBClient.write.

bgusach:
Centralize default parameters in a function, and reuse these defaults in
InfluxDBClient._write_points and .query.


patricktokeeffe:
A better approach would be to collect params inside `write()`, as was done for `write_points()` in https://github.com/influxdata/influxdb-python/commit/7fae65ff0fcabb998b02d22c136edb19ffd4da7f

Also, `write()` should get a new `database=None` argument (like `write_points` has) that documents the use of connection's default database.

Issue #649, opened 2018-10-13
Incorrect nanosecond timestamps being written to influxdb


auphofBSF:
Writing point with time precision  in ns results in incorrect nanosecond timestamps being written to influxdb
effect can be noticed in writing points from both the  `DataFrameClient.writepoints`as well as the `InfluxDBClient.write`
This has been noted in issues: #527,#489, #346 and #344 and #340 

  and a proposed pull request submitted. [https://github.com/influxdata/influxdb-python/pull/346](https://github.com/influxdata/influxdb-python/pull/346) .  I believe this pull request does not go far enough

#489 describes the same "floor divide" solution as part of the solution being proposed here. 

This solution I am proposing in a pull request will use, floor divide and  pandas.Timestamp that supports ns resolution.

### Implication
In trying to use the library to update a record it writes a new record with the erroneous timestamp

### Effected code regions
`_dataframe_client` 
   _convert_dataframe_to_json
   _convert_dataframe_to_lines
   _datetime_to_epoch
`line_protocol.py`  
  _convert_timestamp

### Explanation:
the divide of an (np.int64 or int ) nanosecond timepoint by 1 (ns precision)  is producing errors. it is necessary to use the floor divide operator // which yields the correct (np.int64 or int) value

If an np.int64 is divided (Operator /)  by an int  it yields a float64. When the initial  np.int64 value is large enough the conversion  back to an np.int64 looses nanosecond precision. The same error occurs with standard python int types.

### Example (formulated in a ipython notebook)

    .......... dataframe  is from a query (influxdb measurement having particular field value errors) returned 2 timepoints
    dataframe.index  --->   DatetimeIndex(['2018-05-29 00:53:12.889962156+00:00', '2018-10-03 07:06:36.975643599+00:00'], dtype='datetime64[ns, UTC]', freq=None)
    dataframe.index[1]  ---> Timestamp('2018-10-03 07:06:36.975643599+0000', tz='UTC')
    dataframe.index[1].value  ---> 1538550396975643599
    
    ######    alternative method for timepoint fabrication   #########
    timepoint = pd.Timestamp('2018-10-03 07:06:36.975643599+0000', tz='UTC')
    timepoint.value   ---> 1538550396975643599

    type(timepoint.value)   ---> int
    timepoint.value / 1  ----> 1.5385503969756436e+18
    type(timepoint.value / 1) ----> float
    type(timepoint.value // 1) ----> int
[...]

Issue #527, opened 2017-10-26
Cannot work with nanosecond precision


theodiefenthal:
Influx is designed to work with nanosecond precision. Therefore, the timestamp is stored as a 64 bit unsigned Integer internally (And system max time is kind of Long.MAX in terms of Java speak). 

Python, as we know, has no nanosecond support for it's `datetime` objects.
Numpy's datetime64, on the other side, works nicely with nanosecond precision date objects.

If I now create a dataframe like this:
```
                               value
time                                
1970-01-01 00:00:00.000000001     40
2017-06-01 00:00:00.000000000    100
2017-06-10 13:13:13.123456789    150
```
and call within the DataFrameClient `_convert_dataframe_to_lines`, I'll end up with
```
['myMeasurement value=40i 1', 'myMeasurement value=100i -432406528', 'myMeasurement value=150i 501012992']
```
See how the timestamp is correct only for the first entry.

Another bug appears when using json format instead:
```
[{'measurement': 'myMeasurement', 'tags': {}, 'fields': {'value': 40}, 'time': 1}, {'measurement': 'myMeasurement', 'tags': {}, 'fields': {'value': 100}, 'time': 1496275200000000000}, {'measurement': 'myMeasurement', 'tags': {}, 'fields': {'value': 150}, 'time': 1497100393123456768}]
```
See how for the last value, the last nanosecond precision is truncated. 

In order to write data via DataFrameClient, I sadly have to use a python datetime format and cannot go for an int or something as time column directly due to type checks.

---

In InfluxDBClient docu for write_points, you write that `:param time_precision: Either 's', 'm', 'ms' or 'u', defaults to None` even though in the code, you also allow for `n`. 

I didn't test out the InfluxDBClient directly for writing, but for a short glance on the code (looking at `_convert_timestamp`, especially at `ns = (timestamp - EPOCH).total_seconds() * 1e9`, it doesn't look like it is supporting nanosecond or even something more than second precision either. 

*EDIT* Just to let you know, I have used influxdb==4.1.1 in my requirements.txt file and I'm running the code with Python 3.6.

Issue #489, opened 2017-08-14
Nanoseconds not being written correctly to DB (last few digits get changed)


rthapliyal:
In the write_points function with json protocol and time_precision = 'n'.

There is a line in the code (line # 217 in version 4.1.1 - see snippet below ) that basically takes a datetime value and divided it by 1. However the division is done using float (/) versus integer division (//). This changes the last few digits of the datetime being written. E.g. it the value 1502114400290050489 when divided by 1 using / changes to 1.5021144002900506e+18. When using // and dividing by 1 it remains unchanged.

------------CODE COPY BELOW------------------------------------
        points = [
            {'measurement': measurement,
             'tags': dict(list(tag.items()) + list(tags.items())),
             'fields': rec,
             **_'time': int(ts.value / precision_factor)}_**
            for ts, tag, rec in zip(dataframe.index,
                                    dataframe[tag_columns].to_dict('record'),
                                    dataframe[field_columns].to_dict('record'))

PR #650, opened 2018-10-14
[Fix](#649) changes to support accurate ns timepoints

auphofBSF:
as documented in (#649). 
use floor divided on integer division and use pandas.Timestamp for ns timestamp resolution

Extensively tested but not fully tested, unit test updates to follow
I believe address and fixes ( #527,#489, #346, #344, #340)


aviau:
I'd be happy to merge if it includes unit tests.

auphofBSF:
Have added fixes to enable passing existing python unit tests.  
#### However some failures persist

1. Unit Test failures are with pypy  because of the now mandatory dependency on pandas to do ns precision timestamps. I have no experience with pypy can we allow dependency on pandas for pypy ?
2. The other unit test failure is flake8,  I cannot figure this one out HELP please someone


I am testing further unit test updates and additions for confirmation of ns precision conformance.

tpietruszka:
While I believe this PR improves the situation, I still see a problem.

**When trying to write points with microsecond precision, the timestamps are sometimes erroneous (1 us too early)**

Why? **The `precision_factor` values (other than 1), in `_dataframe_client.py`, are floats, and so the result of floor division is also a float.** For microsecond precision this poses a problem - most timestamps only have approximate float representations. If the approximation is smaller than the actual value, the next call to `astype(np.int64)` takes a floor value of something smaller than the value we want.

**Solution: 
`precision_factor` values should be ints. BTW, maybe they could be defined just once.**



Example:
```
import pandas as pd
import influxdb

test_db = 'test_python_client_timestamp_precision'
col_name = 'temperature'
measurement = 'test'

df = pd.Series({pd.Timestamp('2019-04-16 16:24:56.915728Z'): 1}, name = col_name).to_frame()

client = influxdb.DataFrameClient(database=test_db)
client.create_database(test_db)
client.write_points(df, measurement, time_precision='u')

df2 = client.query(f'select * from {measurement}')[measurement]
client.drop_database(test_db)
client.close()
pd.testing.assert_index_equal(df.index, df2.index)
```
**(the assert fails both in current master AND in a version including this PR)**

auphofBSF:
The issue identified by @tpietruszka  https://github.com/influxdata/influxdb-python/pull/650#issuecomment-485915261 is valid 

It happens on certain values, I am not well experienced on numpy arithmetic particular corner cases of conversion to challenge what is correct or incorrect. But @tpietruszka code and the simple experiment of test values and different arithmetic strategies for conversion recorded below demonstrate the issue.  I have changed the test values in the Unit tests to 

>  '2013-01-01 23:10:55.123456987+00:00' 

This change to unit tests is in the following branch https://github.com/auphofBSF/influxdb-python/tree/bug/microsecond_issue_tpietruszka.
 **Certainly all tests fail.**  


The  @tpietruszka https://github.com/influxdata/influxdb-python/pull/650#issuecomment-485915261 result should be identical (There are no NanoSeconds)  but where there are nanoseconds what to do ?

### A QUESTION and DECISION require about Rounding or Truncating
In the following experiment where there are NanoSecond components of the time should the value be truncated to the MicroSecond or Rounded Up to the Microsecond and for that matter every conversion precision.  

I am happy to assist with further alterations, thanks to @tpietruszka for identifying and solving to this stage

image


```python
ts = pd.Timestamp('2013-01-01 23:10:55.123456987+00:00')
ts
ts_ns = np.int64(ts.value)

ts_ns
expected_ts_us = 1357081855123456

#as before Pull Request #650
result1 = np.int64(ts_ns / precision_factor)
(expected_ts_us == result1, result1)

# as in Pull Request #650
result2 = np.int64(ts_ns // precision_factor)
(expected_ts_us == result2, result2)


# Tests for Fix identified by @tpietruszka  DIVIDE
result3=np.int64(ts_ns / np.int64(precision_factor))
(expected_ts_us == result3, result3)

# Tests for Fix identified by @tpietruszka  FLOOR DIVIDE
result4=np.int64(ts_ns // np.int64(precision_factor))
(expected_ts_us == result4, result4)

```

influxdb/line_protocol.py
```diff
@@ -9,37 +9,45 @@
 from datetime import datetime
 from numbers import Integral
 
-from pytz import UTC
-from dateutil.parser import parse
 from six import iteritems, binary_type, text_type, integer_types, PY2
 
-EPOCH = UTC.localize(datetime.utcfromtimestamp(0))
+import pandas as pd  # Provide for ns timestamps
```
lipi: Please add pandas to requirements.txt (and remove pytz and python-dateutil), build is currently failing because of that, see pypy build logs: https://travis-ci.org/influxdata/influxdb-python/jobs/516866320

Great fix btw, looking forward to being merged.

auphofBSF:
The commits now include fixes for @tpietruszka  identified issues with precision conversion. Test cases have been updated and pass. 

The only failures still exist with pypy.   Following @lipi recommendation in https://github.com/influxdata/influxdb-python/pull/650#discussion_r292698794, enabled pypy3 to pass but then failed on all standard python implementations  clash of  tox.ini deps: and requirements.txt  duplicate requirements. 
Additionally pypy2.7 fails with pandas having a dependency on numpy 1.17 which only supports python >3.5.
I have a seperate local branch https://github.com/auphofBSF/influxdb-python/tree/test/pypy27pandas where I will explore fixing this

@lipi 
1) do you have ANY further suggestions regarding PYPY2.7 and pandas
2) how best to add pandas to requirements and have CI tests still use there depend in tox.ini

clslgrnc:
#407 seemed much more lightweight to solve the problem. I am not sure why panda is needed just for time conversion, and I do not really understand why all the powers of 10 are defined as floats with the notation 1eX (which result in a float in python) then converted to numpy 64 bits ints. Why not just use 10 ** X which is already a python integer with arbitrary precision?

sebito91:
If this still fails `pypy` inclusion I will re-address #407.

sebito91:
Duplicate of #407 without `pandas` requirement

Issue #671, opened 2019-01-07
Pandas 0.23.4 TypeError: Already tz-aware


fl4p:
The UTC[ tz localization](https://github.com/influxdata/influxdb-python/blob/master/influxdb/_dataframe_client.py#L205) raises a `TypeError: Already tz-aware, use tz_convert to convert.` with pandas 0.23.4 . 

I think this a suitable fix:
```
            if df.index.tzinfo is None:
                df.index = df.index.tz_localize('UTC')
```

grainednoise:
I ran into the same problem. The proposed fix solves the problem for me.

nathanielatom:
Any ETA on when this fix will be merged and released?

lovasoa:
I got the same problem. Is something preventing the PR from being merged ?

<details>

```
File "/lib/python2.7/site-packages/influxdb/_dataframe_client.py", line 185, in query
return self._to_dataframe(results, dropna)
File "/lib/python2.7/site-packages/influxdb/_dataframe_client.py", line 205, in _to_dataframe
df.index = df.index.tz_localize('UTC')
File "/lib/python2.7/site-packages/pandas/core/accessor.py", line 91, in f
return self._delegate_method(name, *args, **kwargs)
File "/lib/python2.7/site-packages/pandas/core/indexes/datetimelike.py", line 721, in _delegate_method
result = operator.methodcaller(name, *args, **kwargs)(self._data)
File "/lib/python2.7/site-packages/pandas/core/arrays/datetimes.py", line 1049, in tz_localize
raise TypeError("Already tz-aware, use tz_convert to convert.")
TypeError: Already tz-aware, use tz_convert to convert.
```

</details>

AlnsV:
This solves the problem, happening to me just now...

freshwuzhere:
Me too :-(

lovasoa:
#679 fixes this. @xginn8  should [merge it soon](https://github.com/influxdata/influxdb-python/pull/679#issuecomment-467851424)

clslgrnc:
Should be solved in 5.2.2

PR #674, opened 2019-01-17
Unpin setuptools to fix travis

xginn8:
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>

PR #678, opened 2019-02-07
Parameter binding for client's `query()` method

### LINKED ISSUE #316: [feature][security] Add an option to escape queries
Ideally the the query method would work like that:

``` python
c.query("select value from measurement where tag = :value", value="untrusted input")
```

where value gets escaped and discourage the use of:

``` python
c.query("select value from measurement where tag = %s" % "untrusted input")
```

Reason is that people might use this in their application where values comes in they not control
and we end up with something like xkcd 327: https://xkcd.com/327/
I am not an expert on SQL injection as there are some nasty edge cases, but maybe somebody in the community is. I think, libraries should make it hard for users to make mistakes.
This is not only security relevant. Those stuff can happen also unintentionally as in #309

### LINKED ISSUE #603: Client's `query()` method should directly support parameter binding
The `InfluxDBClient.query()` method does not directly support the parameter binding syntax, documented at [1].  This can be worked around, by passing a `params` dictionary containing the JSON-encoded parameters as the "params" key, but this is rather confusing ("Which 'params' exactly are you talking about?  How do I encode it?").  The `query()` method really should treat bound parameters as first-class citizens in some fashion.  The obvious way of doing this would be to have `query()` accept `**kwargs` and, if it is non-empty, JSON-encode it and add it to the `params` dictionary under the "params" key.  Alternatively, the `query()` method could take a `bind` or `bound` parameter, which must be a dictionary that would be treated the same way.

Either of these would match similar interfaces provided by other Python database libraries and would help eliminate the threat of query language injection, which is a potentially severe class of bugs.

[1] https://docs.influxdata.com/influxdb/v1.3//tools/api/#bind-parameters

clslgrnc:
Closes #603, closes #316. Enables the following:
```python
client.query('select Int_value from cpu_load_short where host=$host_var;',
             bind_params={'host_var': 'server01'})
```

~~The last commit ensures all tests pass and is unrelated to the rest of this PR (feel free to revert it, #671 is probably a better suggestion).~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+- query() now accepts a bind_params argument for parameter binding (#678 thx @clslgrnc)
 
 ### Changed
 
--- a/examples/tutorial.py
+++ b/examples/tutorial.py
@@ -13,7 +13,9 @@ def main(host='localhost', port=8086):
     dbname = 'example'
     dbuser = 'smly'
     dbuser_password = 'my_secret_password'
-    query = 'select value from cpu_load_short;'
+    query = 'select Float_value from cpu_load_short;'
+    query_where = 'select Int_value from cpu_load_short where host=$host;'
+    bind_params = {'host': 'server01'}
     json_body = [
         {
             "measurement": "cpu_load_short",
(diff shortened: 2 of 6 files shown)


clslgrnc:
Obviously, travis fails without #674 and is unable to check this PR properly.

clslgrnc:
Used @fl4p more elegant fix: bcac8072e254eabbfd2451708585bb7540c592bd

clslgrnc:
Rebasing to #679, hoping it will eventually get merged.

clslgrnc:
rebased to master (let me know when rebases stop being acceptable) to remove #679 now that travis is fixed

xginn8:
@clslgrnc can you add a changelog entry please?

clslgrnc:
Done.

xginn8:
lgtm, thank you @clslgrnc!

PR #679, opened 2019-02-13
Fix Travis and Panda tz_localize

clslgrnc:
Travis is still broken.

It should work with the included changes:
- unpin pypy as cryptography-2.5 is not compatible with PyPy < 5.4 (see [failed build](https://travis-ci.org/influxdata/influxdb-python/jobs/492607023))
- apply @f4bsch patch (see [failed build](https://travis-ci.org/influxdata/influxdb-python/jobs/486383231), closes #671 and closes #676)
- ignore N816 `mixedCase variable in global scope`, might not be the best solution but getting travis back up should be a priority (see [failed build](https://travis-ci.org/influxdata/influxdb-python/jobs/492607042)).

PR #683, opened 2019-03-05
ignore N816: mixedCase variable in global scope

clslgrnc:
ignore N816 `mixedCase variable in global scope`, might not be the best solution but getting travis back up should be a priority (see [failed build](https://travis-ci.org/influxdata/influxdb-python/jobs/492607042)).

Extracted from #679

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,7 @@ commands = nosetests -v --with-doctest {posargs}
 deps =
   flake8
   pep8-naming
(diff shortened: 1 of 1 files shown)


xginn8:
@clslgrnc I pushed up a fix to this rather than just an ignore, please take a look if you could: https://github.com/influxdata/influxdb-python/pull/689

clslgrnc:
Sure, it's obviously better than just silencing all triggered alarms :smile:

clslgrnc:
closing as #689 is much better

PR #692, opened 2019-03-15
[WIP] add py37 and recent influxdb

clslgrnc:
+ diversify panda and numpy version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
-  - "3.7"
-    dist: xenial
-    sudo: true
   - "pypy"
   - "pypy3"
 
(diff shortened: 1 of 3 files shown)


clslgrnc:
It seems that a test fails with numpy > 1.13.1
I'll investigate later

clslgrnc:
Here is the culprit:
```
>>> np.version.version
'1.13.3'
>>> a = np.float64('1.1111111111111')
>>> str(a)
'1.11111111111'
```
```
>>> np.version.version
'1.16.2'
>>> a = np.float64('1.1111111111111')
>>> str(a)
'1.1111111111111'
```

clslgrnc:
I really do not know how to put python 3.7 into travis matrix.
An alternative would be to use xenial and sudo:true for all builds. What do you think?

aviau:
> An alternative would be to use xenial and sudo:true for all builds. 

We should avoid doing that. Container builds are scheduled faster (I think) and going back to VMs would be a step backwards.

aviau:
Thank you for the PR!

clslgrnc:
This time you were too fast for me :wink: I pushed the doc and a small fix in #693.
(To your defense I was fiddling with the WIP status of the PR, I felt that travis was not starting fast enough)

PR #693, opened 2019-03-16
Python and influxdb supported versions

### LINKED ISSUE #634: Add python 3.7 support
As python3.7 is out, we need to add python3.7 support.

clslgrnc:
update the doc and fix a small bug when numpy does not have a purely numerical version number.
closes #634


aviau:
Thanks!

PR #702, opened 2019-04-15
Mutual TLS authentication

LloydW93:
A recreation of #677 because I accidentally made a mess of my branch :(

With the Python requests module, it is possible to configure a client certificate to use for mutual TLS authentication, for example if an nginx instance in front of influxdb is performing the TLS termination and has a whitelist of allowed CNs.

This PR adds a new optional argument to the client constructor which passes through the path to such a client certificate to the requests session.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -61,6 +61,11 @@ class InfluxDBClient(object):
     :type proxies: dict
     :param path: path of InfluxDB on the server to connect, defaults to ''
     :type path: str
+    :param cert: Path to client certificate to use for mutual TLS
+        authentication, defaults to None
+    :type cert: str
+
+    :raises ValueError: if cert is provided but ssl is disabled (set to False)
     """
 
     def __init__(self,
@@ -78,6 +83,7 @@ class InfluxDBClient(object):
                  proxies=None,
                  pool_size=10,
                  path='',
+                 cert=None,
                  ):
         """Construct a new InfluxDBClient object."""
         self.__host = host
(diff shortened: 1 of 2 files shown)


petetanton:
We would like to start using mutual TLS with our InfluxDB deployments so that we can evaluate InfluxDB for production use.
@aviau @sebito91 or @xginn8 please can someone review this PR?

influxdb/client.py
```diff
@@ -61,6 +61,11 @@ class InfluxDBClient(object):
     :type proxies: dict
     :param path: path of InfluxDB on the server to connect, defaults to ''
     :type path: str
+    :param cert: Path to client certificate to use for mutual TLS
```
westfeld: I would suggest to be more precise about the fact that this path is pointing to a file containing *both* the client certificate and its private key.

westfeld:
If I got it correctly, for valid client-side certificate authentication you need a client certificate and its matching private key. They can be both in one file or in two separate files.

I would suggest adding this to the documentation of the new parameter like requests does it, that it is also possible to have a tuple with both file paths instead of one path only.

**From the requests documentation**
> You can also specify a local cert to use as client side certificate, as a single file (containing the private key and the certificate) or as a tuple of both files’ paths

[Requests documentation about client certificates](https://2.python-requests.org/en/master/user/advanced/#client-side-certificates)

sebito91:
Sorry, been consumed lately but will get this reviewed this week!

LloydW93:
I've updated the documentation to more explicitly state the supported values. I considered explicitly linking to the Requests documentation - but I'm not sure whether this would be preferred, especially as links can break over time.

kaylashapiro:
Hello! Are there any updates on this PR?

lovasoa:
@aviau I think this PR can be merged.

aviau:
Sure, LGTM.

Issue #705, opened 2019-04-19
ResultSet.get_points: lack tags in their result set


timur-orudzhov:
I think, it would be useful that ResultSet.get_points also should return tags related to query. For example, i have query string:

```python
query_string = (
        'SELECT count("some_field") AS "count" '
        'FROM "measurement" '
        'GROUP BY "some_tag", time(1d)'
    )
```

Wright now, after execution this query i will get only *time* and *some_field_count* fields, but i also need info about "some_tag" and its value.

clslgrnc:
What you are looking for is `items`. As per the [documentation](https://influxdb-python.readthedocs.io/en/latest/api-documentation.html#resultset) `get_points` returns "a generator for all the points that match the given filters" this is done by parsing the raw json and should not be significantly faster than:
```python
for (measurement, tags), points in result_set.items():
    if not my_filter(measurement, tags):
        continue
    for point in points:
        do_something(measurement, tags, point)
```

aviau:
This issue looks like a question more than an actual issue. Closing.

PR #706, opened 2019-04-22
#705 add tags data to result of _get_points_for_series

timur-orudzhov:


clslgrnc:
This would change the behavior of get_points and will cause problems to people that have tags and fields with the same name.

timur-orudzhov:
@clslgrnc Yeah, i was mistaken. Thanks for your clarification!

PR #708, opened 2019-04-25
Docfix retention policy replication type

bbinet:
should be int (not str)

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -660,7 +660,7 @@ class InfluxDBClient(object):
             The minimum retention period is 1 hour.
         :type duration: str
         :param replication: the replication of the retention policy
-        :type replication: str
+        :type replication: int
         :param database: the database for which the retention policy is
             created. Defaults to current client's database
         :type database: str


bbinet:
this PR would also make it consistent with alter_retention_policy replication type: https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L709

influxdb/client.py
```diff
@@ -660,7 +660,7 @@ def create_retention_policy(self, name, duration, replication,
             The minimum retention period is 1 hour.
         :type duration: str
         :param replication: the replication of the retention policy
-        :type replication: str
+        :type replication: int
```
xginn8: Since we don't actually enforce this parameter via any sort of type check and it's just used in a large string format, it can be either a string or an int. Would you mind updating that?

bbinet: I think that "str" type may mislead people as only string representation of actual numbers are actually supported (not arbitrary strings).
I think it is better to set type to int only like for existing alter_retention_policy replication param, so that it makes it clear that only numbers are allowed.

xginn8: Imo this change should be accompanied by an explanation of what the field is or does rather than just changing the type. Additionally, a more complete change might be also type-checking the field and raising an error if it is the wrong type.

sebito91:
Closing stale issue. If we need this in future please let me know and we can re-open.

PR #710, opened 2019-05-03
Added headers parameter to InfluxDBClient

nathanielatom:
Hi, we'd like to use the `DataFrameClient` for writing and querying data from our Influx DB, which is accessible through an API Gateway we have setup. This requires an API key in the header of any requests for authentication. 

This commit simply adds an optional `headers` parameter to `InfluxDBClient.__init__`. The default is {} and if 'Content-Type' or 'Accept' aren't present they'll be added with the same values as before.

Note I also made another very small change in `InfluxDBClient.write` which I think provides a small bug fix, since previously the 'Content-Type' of the `self._headers` dict was getting inadvertently modified.

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -84,8 +84,10 @@ class InfluxDBClient(object):
     :param session: allow for the new client request to use an existing
         requests Session, defaults to None
     :type session: requests.Session
+    :param headers: headers to add to Requests, will add 'Content-Type'
+        and 'Accept' unless these are already present, defaults to {}
+    :type headers: dict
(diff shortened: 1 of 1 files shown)


sebito91:
Thanks @nathanielatom will take a look and get back to you!

nathanielatom:
@sebito91 any thoughts or feedback?

sebito91:
@nathanielatom I'm sorry for the extremely late reply, but I think this is a nice PR. Could you please rebase against master and make sure to take the current defaults there?

nathanielatom:
Sure! :)

Issue #632, opened 2018-08-21
Cryptic error a tag contains a newline


salty-horse:
Using influxdb package version 5.2.0.

When calling `write_points` with a tag string that contains a newline, the client fails on an error that doesn't mention the tag:

```python
from influxdb import InfluxDBClient
json_body = [
    {
        "measurement": "cpu_load_short",
        "tags": {
            "host": "server01",
            "region": "us-west\n" # Note the newline here
        },
        "time": "2009-11-10T23:00:00Z",
        "fields": {
            "value": 0.64
        }
    }
]
client = InfluxDBClient('localhost', 8086, 'root', 'root', 'example')
client.create_database('example')
client.write_points(json_body)

InfluxDBClientError: 400: {"error":"unable to parse 'cpu_load_short,host=server01,region=us-west': missing fields\nunable to parse 'value=0.64 1257894000000000000': invalid field format"}
```
Note that the error doesn't mention the newline at all.

If a newline is legal in a tag name, it may be serializing the command to influx incorrectly. If a newline isn't legal, the error should be clearer.

PR #716, opened 2019-05-16
Fix newline in tag value cause partial commit

syhan:
When a tag value contains newline(\n), the request sent to db would be splitted into two parts and the first part would fail to write to db but the second would be succeed. The reason is that before sending we do serialization(make_lines) the _escape_tag method in line_protocol.py won't handle it well, we need somehow more specific on newline instead of only handling escape character (\)

There have been an issue raised before #632

### DIFF
--- a/influxdb/line_protocol.py
+++ b/influxdb/line_protocol.py
@@ -54,6 +54,8 @@ def _escape_tag(tag):
         ",", "\\,"
     ).replace(
         "=", "\\="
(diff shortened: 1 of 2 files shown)


lovasoa:
@aviau I think this PR is worth merging.

aviau:
Alright, LGTM.

lovasoa:
Great, thank you @aviau !

PR #721, opened 2019-06-05
Implemented the context manager protocol

JustusAdam:
Python supports so called [context managers](https://docs.python.org/3/reference/datamodel.html#context-managers). These can be used to automatically cleanup resources, even in the case of exceptions.

This PR implements the methods necessary for the context manager protocol (`__enter__` and `__exit__`).


aviau:
Okay, this is fine.

Can you write a test please?

JustusAdam:
Ah, whoops, I somehow didn't see you commented on this... Sorry for not responding, and thanks for implementing this :+1:

Issue #300, opened 2016-02-24
Specify Retention Policy in SeriesHelper 


matt-snider:
It would be good to be able to specify the Retention Policy on a per `SeriesHelper` basis.

At the moment I'm employing a straight-forward hack:

``` python
class MySeriesHelper(SeriesHelper):
     class Meta:
        # ....

     @classmethod
     def commit(cls, client=None):
        # Overwrite this method to add the retention policy. This is a temp
        # hack until this can be applied to upstream.
        if not client:
            client = cls._client
        rtn = client.write_points(cls._json_body_(), retention_policy='my_retention_policy')
        cls._reset_()
        return rtn
```

I'm not sure if I'm missing something and there is actually a better way to do this. But if not, I think it can be fairly easily implemented by adding a new parameter to `Meta`.

I'd be interested in working on this if you think it's a worthwhile feature.

sebito91:
Addressed in #723

PR #723, opened 2019-06-08
Specify Retention Policy in SeriesHelper

csanz91:
I have added an option in the SeriesHelper meta options to be able to write the data with a custom retention policy.

This request was previously commented in this issue: #300


xginn8:
Hi @csanz91 please ensure that all the tests are passing in Travis, and then we can take a look

csanz91:
All tests are passing now. Feel free to review the pull request.

xginn8:
Hi @csanz91 and thanks for submitting this patch! The implementation looks fine, but please add some tests to ensure this feature functions properly. You'll want to look here: https://github.com/influxdata/influxdb-python/blob/master/influxdb/tests/helper_test.py

csanz91:
I have added a couple of tests in which I check the args passed to the "write_points" function to look for the specified retention policy. I also have a case without a retention policy to make sure I'm not altering the default rp.

Issue #718, opened 2019-05-26
gzip support on influxdb-python


philip-jones:
Hi,

Great API, we use it heavily.  Thank you.

The equivalent Java API influxdb-java, which I also use, allows gzip compression to be explicitly enabled when setting up a connection, e.g.

```java
final javaInfluxClient = InfluxDBFactory.connect(SERVER, USERNAME, PASSWORD);
javaInfluxClient.enableGzip();
// ...etc
```
Perhaps this is enabled by default in influxdb-python, however I cannot find any reference to gzip compression.  Perhaps I am looking in the wrong place.

This would be very useful to allow a reduction in bandwidth for IoT devices with very small data tariffs.

Would it be possible to add this?  I'm happy to have a go at writing a patch for this but would like to get some feedback from the maintainers first.

Thank you, Phil.

driftregion:
Related: https://github.com/influxdata/influxdb-python/issues/451
In Oct 2017, xginn8 mentioned that there was no InfluxDB API documentation for gzip compression.  Perhaps this has changed since then?
https://docs.influxdata.com/influxdb/v1.7/guides/writing_data/#configure-gzip-compression

KEClaytor:
Hello, I just wanted to jump in and let you know that I've been able to add gzip compression to the client and have made a PR #732 

I agree that data compression is really important, especially on IoT devices and mobile networks.

I'm unsure how to properly do the tests so it may still be awhile until it's merged.

philip-jones:
Thank you for the response KEClaytor & driftregion, much appreciated.

Will be happy to do some testing when you are ready / there is a consensus about how this should be done.

lovasoa:
Related: #733 MessagePack greatly reduces network traffic between the client and the server.

KEClaytor:
@driftregion Yes, compression seems to be supported. Thanks for linking to the official docs on this. The earliest mention of it in the changelog is in v1.0:
https://docs.influxdata.com/influxdb/v1.7/about_the_project/releasenotes-changelog/

@lovasoa Message Pack looks really neat. The only reference I can find to it in the docs is:
 - Add message pack format for query responses.

From: https://docs.influxdata.com/influxdb/v1.7/about_the_project/releasenotes-changelog/

[...]

PR #732, opened 2019-07-09
gzip compression for data (post and responses)

KEClaytor:
Hello, I have enabled gzip content encoding in this branch. I have tested it against my own influx application and it seems to work. However, I am unsure how to properly do the tests.

I have included some tests, however, they do not actually track integration with the influxdb application. For instance, the tests pass when I use zlib do do the compression, even though that does not work in my application. Can you advise where / how to do proper implementation tests?

If you would like to incorporate it into the main branch, I would certainly find it useful and others may as well (see: #718).


KEClaytor:
Okay, figured out zlib so that it's compatible with python 2.7.
Local tests seem to pass, but I'm not testing against a range of influxdb versions (only 1.7).
I basically duplicated all server tests but enabled gzip=True.
There are no tests for influxdb08 - are they needed?

influxdb/client.py
```diff
@@ -66,8 +67,9 @@ class InfluxDBClient(object):
         as a single file containing the private key and the certificate, or as
         a tuple of both files’ paths, defaults to None
     :type cert: str
-
     :raises ValueError: if cert is provided but ssl is disabled (set to False)
+    :parm gzip: use gzip content encoding to compress requests
```
xginn8: this should be `param`, and should come before the `:raises` block for consistency

KEClaytor: No problem, thanks!

KEClaytor:
Interestingly I noticed that the zlib compressed data wasn't actually being added to InfluxDB. Switching back to gzip and manually ensuring that data is posted.

f0rtung:
Is there anyone alive?

cgiraldo:
Any chance to see this PR merged?

PR #737, opened 2019-07-18
Fixing tutorial

Aeium:
The tutorial did not work for me with the protocol set to json.
It ran without errors but nothing was written to the database.
It worked as expected after changing the protocol to line.
Banged my head against a wall here for a while, would like to save others the trouble.

### DIFF
--- a/examples/tutorial_pandas.py
+++ b/examples/tutorial_pandas.py
@@ -12,7 +12,7 @@ def main(host='localhost', port=8086):
     user = 'root'
     password = 'root'
     dbname = 'demo'
-    protocol = 'json'
+    protocol = 'line'
 
     client = DataFrameClient(host, port, user, password, dbname)
 


sebito91:
Thanks for the commit @Aeium! Can you please help me understand what issue you're having with the `json` protocol in the tutorial you've referenced?

1. Which version of pandas + python are you using?
2. Which version of influxdb are you using?
3. Do you have any logs from influxdb that you can share?

Thanks!

sebito91:
I think I'm seeing the bug...

https://github.com/influxdata/influxdb-python/blob/dc83fc6576b6463dcc77a0c101475a2a71ed655a/influxdb/_dataframe_client.py#L273-L281

The `tag_columns` list is empty in both examples, and since `zip` stops when the shortest element is [exhausted](https://docs.python.org/3.7/library/functions.html#zip) we end up with an empty object at the end of it.

Another thing I've noticed is that all of the `dataframe` tests are using `line` protocol, so there are a couple of issues we'll need to address to fix this. 

Thanks again for pointing this out!

PR #741, opened 2019-07-29
Types matching

iakrevetkho:
Add function into **write_points** to get **fields types** of measurement, before downloading, and if field exists, convert to existed type.
This feature is very usefull if you don't care about typing your date, when download and can will avoid error with types mismatch if you try to load **float** into **integer** field.

P.S. As bonus I add VSCode files into gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ docs/build/
 .coverage
 cover
 env
+.vscode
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -21,3 +21,4 @@ docs/build/
 .coverage
 cover
 env
+.vscode
```
aviau: This should be in your system's global .gitignore.

It does not make sense to add every single file created by every single editor in the world to every project's gitignore.

aviau:
Thanks for contributing!

Please add a test so that we consider to merge the patch.

Also, please make it optional as it could bread existing uses of the library.

.gitignore
iakrevetkho: Okey, thanks for link!

PR #746, opened 2019-08-12
Make batched writing support all iterables

JayH5:
We have some data we write to InfluxDB from lazy iterables/generators that don't necessarily support indexing or have a length that can be known up-front.

This is a second attempt at #408. It includes a fix for Python 3.7 compatibility (catches the `StopIteration`).

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -6,9 +6,9 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import itertools
 import time
 import random
+from itertools import chain, islice
 
 import json
 import socket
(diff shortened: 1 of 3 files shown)


JayH5:
Build is now green so I'd appreciate a review 🙏

influxdb/client.py
```diff
@@ -6,14 +6,14 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import itertools
```
xginn8: Rather than importing the whole `itertools` module, how about just importing what's needed (`islice` & `chain`)?

influxdb/client.py
```diff
@@ -8,12 +8,12 @@
 
 import time
 import random
+from itertools import chain, islice
```
JayH5: These imports don't seem to be in a particular order.

JayH5:
@xginn8 done, thanks.

JayH5:
@xginn8 maybe we can get this merged now?

PR #753, opened 2019-09-08
Fix chunked query to return chunk resultsets

### LINKED ISSUE #531: Out of memory error when querying large data set with chunked=True

hrbonz:
When querying large data sets, it's vital to get a chunked responses to manage memory usage. Wrapping the query response in a generator and streaming the request provides the desired result.
It also fixes `InfluxDBClient.query()` behavior for chunked queries that is currently not working according to [specs](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py#L429)

close #585
close #531
close #538

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -231,7 +231,7 @@ class InfluxDBClient(object):
         self._username = username
         self._password = password
 
-    def request(self, url, method='GET', params=None, data=None,
+    def request(self, url, method='GET', params=None, data=None, stream=False,
                 expected_response_code=200, headers=None):
         """Make a HTTP request to the InfluxDB API.
 
@@ -243,6 +243,8 @@ class InfluxDBClient(object):
         :type params: dict
         :param data: the data of the request, defaults to None
         :type data: str
+        :param stream: True if a query uses chunked responses
+        :type stream: bool
         :param expected_response_code: the expected response code of
             the request, defaults to 200
         :type expected_response_code: int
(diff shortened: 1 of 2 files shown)


hrbonz:
I'm currently running tox tests and will fix whatever fails in travisCI

psy0rz:
Does this also close https://github.com/influxdata/influxdb-python/pull/538 ?

hrbonz:
Yes, it covers it as well. I've had issues with my laptop so got caught in a bit of a snag, will get back to fixing the tests.
The tests were designed to pass rather than reflect the method documented behavior.

hrbonz:
I had to change the test substantially as it was testing a broken behavior not the documented (and better one imho).
I'm testing that I'm getting the proper number of chunked `ResultSet`s and testing that each `Resultset` returns the proper elements.

hrbonz:
On a side note, I've been using this code to move data from an old version of influxdb to another (and doing some type casting in the middle) with no issue. About 50G of data across 1500 measurements.

russorat:
@hrbonz Thank you!  i've reached out to @sebito91 to take a look.

hrbonz:
Thanks @russorat, hopefully this PR is getting some attention while the new library is getting all the love ;)

hrbonz:
@sebito91 quite happy to see it finally done and merged! Now to 

ping @psy0rz for the inspiration

psy0rz:
awesome! thank YOU for actually implementing it all the way :)

PR #771, opened 2019-12-03
fix: resolving raising exception when calling the commit function without any SeriesHelper

ocworld:
If any SeriesHelper is not generated, calling the commit function makes raising exception because _datapoints is None.

It is hard to find the reason of this error. it is because reviewing influxdb-python's code is needed.

So, I've fixed that if _datapoints is None in the commit function, _datapoints is set to defaultdict to avoid raising error.

---
##### Contributor checklist

- [x] Builds are passing
=> it is not feature addition but covering an exception.


ocworld:
I'll request this pull request again with small changes.

PR #772, opened 2019-12-05
fix: Calling commit and _json_body make raising an exception when any datapoints are not added.

ocworld:
If any SerialHelper is not generated, calling the commit function makes raising an exception because _datapoints is not allocated.

It is hard to find the reason of this error. it is because reviewing influxdb-python's code is needed.

I think that it is important that is producing predictable results. Results when calling first time is needed to equal to calling resetting datapoints in json_body.

So, I've fixed that if not initialized when calling _json_body() function, _datapoints is reset to avoid raising error.

In Unittest, the setup function is added. When calling setup function firstly, __initialized__ is False and _datapoints is not assigned. But, because of this commit, it is OK.

Contacts: Keunhyun Oh <keunhyun.oh@ahnlab.com>

---
##### Contributor checklist

- [x] Builds are passing

### DIFF
--- a/influxdb/helper.py
+++ b/influxdb/helper.py
@@ -154,6 +154,8 @@ class SeriesHelper(object):
         :return: JSON body of these datapoints.
         """
         json = []
+        if not cls.__initialized__:
+            cls._reset_()
         for series_name, data in six.iteritems(cls._datapoints):
             for point in data:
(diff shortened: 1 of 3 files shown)


ocworld:
@Ahnlab-oss

PR #783, opened 2020-02-04
Fixes/msgpack default

jgspiro:
### msgpack issue
I identified an issue with the move to the default use of **msgpack**, so added this pull request to revert to the old behaviour, and optionally enable msgpack by setting `use_msgpack=True` in the __init__ function of the client.

The test `test_write_points_mixed_type` fails on purpose now because of a difference between the json and msgpack result returned by the server.  The question is : is moving to msgpack going to be a breaking change?  If so, I propose to merge this pull request (after fixing the test) and let users manually upgrade to msgpack based parsing.

This pull request has a couple of other changes identified while getting to the bottom of the above issue.

### Fixes
- raise proper exception when send_packet is called with invalid protocol (accessing uninitialzed member)
- docstring : fix missing 'optional' flags for query() parameters
- fix shadowed parameters response and filter
- client_test : some tests using @raises(Exception) were succeeding **for the wrong reason**
(assertion in mock function): using self.assertRaises() instead.

### Others
- test all client functions with both msgpack enabled and disabled
- add extra msgpack test starting from json with int, float and string values
- retention policy creation : cast **replication** to int, adjust docs

---
##### Contributor checklist

- [ ] Builds are passing --> waiting for feedback on msgpack issue

---
##### Contributor checklist

- [ ] Builds are passing

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -89,6 +89,7 @@ class InfluxDBClient(object):
                  pool_size=10,
                  path='',
                  cert=None,
+                 use_msgpack=False
                  ):
         """Construct a new InfluxDBClient object."""
         self.__host = host
@@ -110,7 +111,9 @@ class InfluxDBClient(object):
         )
 
         if use_udp:
-            self.udp_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            self._udp_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        else:
+            self._udp_socket = None
 
         if not path:
             self.__path = ''
@@ -145,10 +148,16 @@ class InfluxDBClient(object):
             self._port,
             self._path)
 
-        self._headers = {
-            'Content-Type': 'application/json',
-            'Accept': 'application/x-msgpack'
-        }
+        if use_msgpack:
+            self._headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'application/x-msgpack'
+            }
+        else:
+            self._headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'text/plain'
+            }
 
     @property
     def _baseurl(self):
@@ -243,14 +252,14 @@ class InfluxDBClient(object):
         :param method: the HTTP method for the request, defaults to GET
         :type method: str
         :param params: additional parameters for the request, defaults to None
-        :type params: dict
+        :type params: dict, optional
         :param data: the data of the request, defaults to None
-        :type data: str
+        :type data: str, optional
         :param expected_response_code: the expected response code of
             the request, defaults to 200
         :type expected_response_code: int
         :param headers: headers to add to the request
-        :type headers: dict
+        :type headers: dict, optional
         :returns: the response from the request
         :rtype: :class:`requests.Response`
         :raises InfluxDBServerError: if the response code is any server error
@@ -285,6 +294,7 @@ class InfluxDBClient(object):
                     verify=self._verify_ssl,
                     timeout=self._timeout
                 )
+                response._msgpack = None
                 break
             except (requests.exceptions.ConnectionError,
                     requests.exceptions.HTTPError,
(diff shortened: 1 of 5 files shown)


influxdb/client.py
```diff
@@ -145,10 +147,16 @@ def __init__(self,
...
-        }
+        if use_msgpack:
+            self._headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'application/x-msgpack'
+            }
+        else:
+            self._headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'text/plain'
```
lovasoa: Why `text/plain` and not `application/json` ?

lovasoa:
> I identified an issue with the move to the default use of msgpack
What is this issue you identified ?

influxdb/tests/server_tests/client_test_with_server.py
```diff
@@ -305,6 +358,46 @@ def test_write_points_DF(self):
...
 
+    def test_write_points_mixed_type(self):
+        """Test writing points with mixed type and reading them in again."""
+        self.assertIs(True,
+                      self.cli.write_points(mixed_series_dummy_points_part_1))
+        self.assertIs(True,
+                      self.cli.write_points(mixed_series_dummy_points_part_2))
+
+        client_1 = InfluxDBClient('localhost', self.influxd_inst.http_port,
+                                  'root', '', database='db', use_msgpack=False)
```
lovasoa: I don't think this test should be added. There is a bug in influxdb (https://github.com/influxdata/influxdb/issues/8707) that causes the JSON response to be parsed as the wrong datatype.

influxdb/client.py
jgspiro: You would have to ask the original implementor - this is the way it was implemented before moving the default to msgpack.

influxdb/tests/server_tests/client_test_with_server.py
jgspiro: Like I said in the description : 

> The test test_write_points_mixed_type fails on purpose now because of a difference between the json and msgpack result returned by the server. 

I can remove the defaulting to json and just make sure all tests run with both json and msgpack client.

I think users would want to be made aware of the consequences of defaulting to msgpack.

lovasoa: The consequences you are talking about is a bug that is now fixed with floats being misinterpreted as integers. I agree with you that this bugfix should be mentioned is the change log. 

In any case, you don't want to have `test_write_points_mixed_type` run with the json client, because it fails due to a bug in influxdb, not in this library.

influxdb/tests/server_tests/client_test_with_server.py
```diff
@@ -305,6 +358,65 @@ def test_write_points_DF(self):
...
+                 "region": "us-west"},
+                {'value': 0.65,
+                 'time': '2009-11-10T23:01:00Z',
+                 "host": "server01",
+                 "region": "us-west"},
+                {'value': 0.35,
+                 'time': '2009-11-10T23:02:00Z',
+                 "host": "server01",
+                 "region": "us-west"},
+                {'value': 1,
```
lovasoa: The fact that influxdb returns 1 instead of 1.0 here is a bug. I don't think you should assert that the bug is present. A future version of influxdb could fix that bug, and you wouldn't want the tests too break then. I think you should just remove this assertion.

jgspiro: The issue you mentioned was closed without a fix, so I doubt this will change soon.  It is a result of the default json marshaling in GoLang (they could probably work around it by customizing).

In contrast to you I believe it would be a good thing that the test would fail if the expected result would suddenly change (since the influx versions used to test are pinned it should not).

lovasoa: > it would be a good thing that the test would fail if the expected result would suddenly change 

Yes, but the expected result in this case is `1.0`, not `1`, since the database contains a float. The problem is that the library does not return the expected result. I don't think this buggy behavior should be enforced by a test.

jgspiro: Both are the same `Number` in JSON, which is the reason the issue was closed in InfluxDB.
I **did** remove the JSON result from the test - altough I still think it should be there, because it is a way to document the current and future behaviour.

lovasoa: Yes, both are the same `Number` in JSON, but that does not make this less of a bug 😛 
And it is indeed a good idea to document the problematic behavior. Maybe you could add something to the doc comments ?

jgspiro: I noticed I forgot to add the use_msgpack param docstring to InfluxDBClient doc, so I just pushed a commit documenting that paramter.

Is there a specific place you want me to describe the msgpack/Json difference?

lovasoa: The doc string of the influxdb client seems like a good place to me.
Also, I'm not sure it's clear: I do not have commit rights on this repository. None of the persons with commit rights are active on the repository anymore.

PR #792, opened 2020-03-01
Support compressed write

noradtux:
Modify InfluxDBClient.write() to allow compression of data before posting. Works for me.
---
##### Contributor checklist

- [x] Builds are passing

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -100,12 +100,13 @@ class InfluxDBClient(object):
         self._database = database
         self._timeout = timeout
         self._retries = retries
-        if compression == True:
+        if compression is True:
             self._compression = 9
         else:
             self._compression = compression
 
-        assert self._compression == False or (self._compression >= 0 and self._compression <= 9)
+        assert self._compression is False or \
+                (self._compression >= 0 and self._compression <= 9)
 
         self._verify_ssl = verify_ssl
 
@@ -362,7 +363,7 @@ class InfluxDBClient(object):
                 data = [data]
             data = ('\n'.join(data) + '\n').encode('utf-8')
 
-        if self._compression != False:
+        if self._compression is not False:
             data = gzip.compress(data, self._compression)
             headers['Content-Encoding'] = 'gzip'
 
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -18,6 +18,7 @@ import requests
 import requests.exceptions
 from six.moves import xrange
 from six.moves.urllib.parse import urlparse
+import gzip
 
 from influxdb.line_protocol import make_lines, quote_ident, quote_literal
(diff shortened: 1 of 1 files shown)


influxdb/client.py
```diff
@@ -98,6 +100,13 @@ def __init__(self,
         self._database = database
         self._timeout = timeout
         self._retries = retries
+        if compression is True:
```
sebito91, line 103: This is confusing as we don't know what type the `compression` parameter should be. Can you please update the function to include comments on what to expect from this paramater? Should `self._compression` be an int, bool, something else?

influxdb/client.py
```diff
@@ -98,6 +100,13 @@ def __init__(self,
...
+            self._compression = 9
+        else:
+            self._compression = compression
+
+        assert self._compression is False or \
```
sebito91, line 108: Why should we blow up here if we don't match one of the values? This should be set to a proper default and move along without any issue for the user.

noradtux: Hmm, ok, will do. Usualy I prefer my code to blow up instead of hiding bugs ;)

influxdb/client.py
noradtux: Well, actually it might be int or bool at that point. If it just True the code will choose a default int value. Seemed sensible to me back when I implemented it. Now I am not so sure anymore ... Suggestion?

noradtux:
> Could you please ... add a test case to ensure this is working as expected?
I guess I need to modify tests/client_test.py? I will try ...

influxdb/client.py
sebito91: I think if anything you should define two vars, one for the switch on/off of compression and another for the level.

influxdb/client.py
sebito91: If you flip this to be two parameters as outline above then this should give you enough assurance that things are working as expected.

sebito91:
If you cannot mock out a test for data, just do a simple test to instantiate the client with your options defined and add your `assert` to confirm they're configured as you'd expect.

I think if anything ensuring the `gzip` header is present in a mocked request should suffice.

cgiraldo:
I think this is already implemented in #732

sebito91:
I think the core if it is indeed implemented in #732, but I saw this as giving users the ability to define the compression level itself.

@noradtux if you feel that this is already covered in #732 and you'd like to rebase + amend to fit that structure, then please continue. Otherwise I'm OK if we decide to close this one outright.

noradtux:
Sorry, didn't have much time to look into this properly. I guess my implementation requires at least python 3.2, so the implementation in #732 seems to be more portable. Hence that one has already been merged this one is redundant and can be closed.

PR #818, opened 2020-04-11
remove msgpack pin

prometheanfire:
The hard lock prevents this from being co-installed with many other
packages.  For instance, it's preventing it from being included in
openstack (which is on 0.6.2 and working on 1.0.0 now).

Signed-off-by: Matthew Thode <mthode@mthode.org>

---
##### Contributor checklist
- [ ] Builds are passing
- [×] New tests have been added (for feature additions)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ python-dateutil>=2.6.0
 pytz
 requests>=2.17.0
 six>=1.10.0
-msgpack==0.6.1
+msgpack


cdessez:
Thanks for merging this fix!

Could we get a new `5.3.1` release for this? Currently the latest release of `influxdb` does not contain the fix and I have to pin it to get my CI pipeline to pass.

Cheers

henningway:
Thanks for the fix! +1 on tagging a version. We get constant pip error messages because of the version conflict.

nemesisdesign:
+1 for a new minor release to fix this

PR #734, opened 2019-07-10
Add support for messagepack

lovasoa:
### Description
This pull request implements fetching data from InfluxDB using msgpack serialization instead of JSON. This fixes issues related to serialization ambiguities in JSON, and makes database queries faster. It does not break compatibility with older influxdb versions that do not support msgpack.

### Related issues
Closes #733
Fixes #665 
Fixes #715  
Fixes #625

### Performance
Here is a performance comparison before (JSON) and after (msgpack) this pull request, for a simple `SELECT *` request on a measurement with 2 tags and three fields of types: integer, float and string:

performance comparison

Using msgpack gives a ~25% speedup.

PR #846, opened 2020-08-23
Fix unpacking of negative timestamps from msgpack

bitoffdev:
In #734, support was added to this python client for messagepack, but it does not match the serialization from influxdb for negative timestamps, because it interprets the timestamps as unsigned.

Influxdb uses the tinylib/msgp library to serialize timestamps. Here is an extract of how the timestamps are serialized.

> Time is encoded as Unix time, which means that location (time zone) data is removed from the object.  The encoded object itself is 12 bytes: 8 bytes for a big-endian 64-bit integer denoting seconds elapsed since "zero" Unix time, followed by 4 bytes for a big-endian 32-bit signed integer denoting the nanosecond offset of the time. This encoding is intended to ease portability across languages. [msgp/write.go](https://github.com/tinylib/msgp/blob/f69a9cf4ee09446ff747e5a65fce943b3786e9b4/msgp/write.go#L594)

**This patch updates the python client to correctly use signed second and nanosecond fields.**

---
##### Contributor checklist

- [x] Builds are passing
  - The only failing test is coverage, which is already failing on master. Fixing the existing issue with coverage is out of scope for this pull request.
  - I modified an existing test to ensure that negative timestamps are parsed correctly when we get a msgpack response.

---

## Extra Manual Verification

As a sanity check, I also tested the following script against influxdb (via Docker) before and after my changes, and the result was as expected: negative timestamps fail before this patch and work correctly following this patch.

```python
from influxdb import InfluxDBClient
json_body = [
    {
        "measurement": "cpu_load_short",
        "tags": {
            "host": "server01",
            "region": "us-west"
        },
        "time": "1960-11-10T23:00:12.12345Z",
        "fields": {
            "value": 0.64
        }
    }
]
client = InfluxDBClient('localhost', 8086, 'root', 'root', 'example')
client.drop_database('example')
client.create_database('example')
client.write_points(json_body)
result = client.query('select value from cpu_load_short;')
assert next(result['cpu_load_short']) == {'time': '1960-11-10T23:00:12.123450Z', 'value': 0.64}
print('Negative timestamps were handled correctly!')
```


bitoffdev:
I rebased this PR onto master, which fixed the failing code coverage test. @sebito91, would you be able to take a look at this? My apologies for the at-mention, but I haven't been able to get any eyes on this patch. Let me know if there is a better way to request reviews for this project. Thanks!

PR #850, opened 2020-09-14
Allow skipping DataFrameClient import

jdoyle93:
Importing the DataFrameClient can take quite a while,
especially when the source files aren't yet loaded into
the page cache.

This patch adds an environment variable, INFLUXDB_NO_DATAFRAME_CLIENT,
to allow users of this package to skip this import in cases where
they don't need it.

---
##### Contributor checklist

- [ ] Builds are passing

### DIFF
--- a/influxdb/__init__.py
+++ b/influxdb/__init__.py
@@ -6,16 +6,20 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import os
+
 from .client import InfluxDBClient
-from .dataframe_client import DataFrameClient
 from .helper import SeriesHelper
 
 
 __all__ = [
     'InfluxDBClient',
-    'DataFrameClient',
     'SeriesHelper',
 ]
+
+if "INFLUXDB_NO_DATAFRAME_CLIENT" not in os.environ:
+   from .dataframe_client import DataFrameClient
+   __all__.append( "DataFrameClient" )
 
 
 __version__ = '5.3.0'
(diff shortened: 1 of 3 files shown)


influxdb/__init__.py
```diff
@@ -6,16 +6,20 @@
...
 from .helper import SeriesHelper
 
 
 __all__ = [
     'InfluxDBClient',
-    'DataFrameClient',
     'SeriesHelper',
 ]
 
+if "INFLUXDB_NO_DATAFRAME_CLIENT" not in os.environ:
```
mhall119, line 20: This should probably be:
```
if "INFLUXDB_NO_DATAFRAME_CLIENT" not in os.environ or not os.environ["INFLUXDB_NO_DATAFRAME_CLIENT"]:
```

To allow for false-values to be used, like INFLUXDB_NO_DATAFRAME_CLIENT=0

mhall119 (changes requested):

influxdb/__init__.py
jdoyle93: Unfortunately, since `os.environ` stores values as strings, `not os.environ["INFLUXDB_NO_DATAFRAME_CLIENT"]` will not be enough to check for falsey string values.

We could add a list of values that we count as falsey and check if the value is in there if you'd like to enable this kind of behavior. Something like:

```if os.environ.get("INFLUXDB_NO_DATAFRAME_CLIENT", "0").lower() not in ["0", "false", ""]``

mhall119: That would work, and I think the false values in your example should be sufficient

jdoyle93:
Builds are now passing (except for coverage, which is broken for unrelated reasons)

PR #851, opened 2020-10-05
feat: add support for custom authorization token

bednar:
Following code should produce request with header: `Authorization=my-token`:

```python
client = InfluxDBClient(username=None, 
                        password=None,
                        headers={"Authorization": "my-token"})

version = client.ping()
```

but current implementation always use HTTP basic auth even if an `username` and  a `password` are not specified:
https://github.com/influxdata/influxdb-python/blob/fc0235e2c07abf09eff4de7f6396a17aca1080f1/influxdb/client.py#L334

If the InfluxDB is behind a proxy, the client cannot connect into InfluxDB since both use the Authorization header.

---
##### Contributor checklist

- [x] Builds are passing

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -31,3 +31,9 @@ Tutorials - UDP
 
 .. literalinclude:: ../../examples/tutorial_udp.py
    :language: python
+
+Tutorials - Authorization by Token
+===============
+
+.. literalinclude:: ../../examples/tutorial_authorization.py
+   :language: python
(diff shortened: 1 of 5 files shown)


rhajek (approved):

Frankkkkk:
Hi @bednar,
any idea why the token must be passed to a `header` param, instead of creating an optional param `token` in the constructor ?
Thanks

bednar:
Hi @Frankkkkk,

There is no specific reason, it just hasn't been implemented yet. This PR only fixes the functioning if `username` and `password` are not defined.

Regards

PR #901, opened 2021-09-08
Prevent overwriting Authorization with basic auth

zgy730:
---
##### Contributor checklist

- [ ] Builds are passing

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -336,7 +336,10 @@ class InfluxDBClient(object):
         _try = 0
         while retry:
             try:
-                auth = (self._username, self._password)
+                if "Authorization" in headers:
+                    auth = (None, None)
+                else:
+                    auth = (self._username, self._password)
                 response = self._session.request(
                     method=method,
                     url=url,


aviau:
Looks good, thanks!

fetzerch:
@aviau Is there a chance that you publish a new patch release to PyPI with this included any time soon?

The patch enables the connection to InfluxDB v2 via:

```
client = influxdb.InfluxDBClient(
        database="xyz",
        headers={
            "Authorization": "Token <base64 encoded token>"
        },
    )
```

This seems to be the only way of running existing scripts (that use InfluxQL queries) against the v1 compatibility API of InfluxDB v2. Unfortunately I do still have some scripts which I can't migrate to Flux because that's still missing some features.

SGStino:
ran into this issue, turns out that username='root', password='root' makes it overwrite the Authorization header

using this fixes it:
```
client = influxdb.DataFrameClient(host='<hostname>', username=None, password=None, headers={'Authorization':'Token <token_without_extra_encoding>'})
```

In version 5.3.1 from pip (which, by the way, doesn't have a tag in this repo yet?)

PR #919, opened 2022-07-13
Unquote username & password when parsing a DSN

dufferzafar:
A DSN like: `influxdb://user:sec#ret@host:port`
should be written as `influxdb://user:sec%23ret@host:port`

RFC 3986 allows characters like # ? to appear in
username / password which should be properly unquoted

Please see relevant Python bug:
https://bugs.python.org/issue18140#msg375109

### DIFF
--- a/influxdb/client.py
+++ b/influxdb/client.py
@@ -21,7 +21,7 @@ import msgpack
 import requests
 import requests.exceptions
 from requests.adapters import HTTPAdapter
-from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


dufferzafar:
SQLAlchemy already handles this: https://github.com/sqlalchemy/sqlalchemy/blob/6e538cf64adfef7f604e3b0d976405a27b8e3d27/lib/sqlalchemy/engine/url.py#L834-L838

PR #922, opened 2022-09-05
Marshal deprecated pandas.util.testing import.

bcrant:
Fixes warning generated by tests.

Attempts to import from pandas public API for newer versions of pandas...
`from pandas.testing import assert_frame_equal`

...resolves to the existing path when not found... 
`from pandas.util.testing import assert_frame_equal` when not found

```
FutureWarning: pandas.util.testing is deprecated. Use the functions in the public API at pandas.testing instead.
    from pandas.util.testing import assert_frame_equal
````

---
##### Contributor checklist

- [ ] Builds are passing

### DIFF
--- a/influxdb/tests/dataframe_client_test.py
+++ b/influxdb/tests/dataframe_client_test.py
@@ -20,7 +20,10 @@ from .client_test import _mocked_session
 
 if not using_pypy:
     import pandas as pd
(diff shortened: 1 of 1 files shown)


bcrant:
@aviau @xginn8 Any objections to this? Can we merge?

